PR #99, opened 2026-07-01
fix(mcp): depgraph_status risk_flags state-specific matching blocked_by

### LINKED ISSUE #97: depgraph_status: make risk_flags state-specific
## Context
Raised in the GitHub Copilot review of #90 (`topos/mcp/tools/depgraph.py`, `_status_to_result`).

## Problem
`topos_depgraph_status` sets `risk_flags = ["composable_unavailable"]` for all non-`PRESENT` states, while `blocked_by` is state-specific (`stale_gitnexus_dir`, `gitnexus_load_error`, `gitnexus_schema_mismatch`, ...). Clients that branch on `risk_flags` alone can't tell the states apart.

Note: the specific signal is **not lost** — it's carried in `blocked_by` — so this is a consistency/ergonomics polish, not a data-loss bug (Copilot's wording slightly overstated it).

## Proposed fix
In `_status_to_result` (`topos/mcp/tools/depgraph.py:~156`), mirror the state-specific `blocked_code` into `risk_flags` alongside `composable_unavailable`. Roughly a one-liner.

## Notes
- Priority: low. Good first issue.

harshjainnn:
Fixes #97

## Problem
`_status_to_result` in `topos/mcp/tools/depgraph.py` set `risk_flags = ["composable_unavailable"]` for all non-PRESENT states, while `blocked_by` carried the state-specific code (`stale_gitnexus_dir`, `gitnexus_load_error`, `gitnexus_schema_mismatch`, etc.). Clients branching on `risk_flags` alone couldn't distinguish which state caused the failure — only `blocked_by` had that detail.

## Fix
Mirror the state-specific `blocked_code` into `risk_flags` alongside `composable_unavailable`, so `risk_flags` now reads `["composable_unavailable", blocked_code]` for any non-PRESENT state (empty list when PRESENT). No data was actually lost before this change (the signal was still in `blocked_by`), so this is a consistency/ergonomics fix rather than a fix for missing data.

## Changes
- Simplified the conditional list-building logic in `_status_to_result` into a single expression that includes both `composable_unavailable` and the state-specific `blocked_code` in `risk_flags`.

### DIFF
--- a/topos/mcp/tools/depgraph.py
+++ b/topos/mcp/tools/depgraph.py
@@ -160,14 +160,11 @@ def _status_to_result(status: DepgraphStatus) -> ToolResult:
     state = DepgraphState(status.state)
     action, next_tool, blocked_code = _STATE_GUIDANCE[state]
     blocked_by = [blocked_code] if blocked_code else []
-    # Mirror the state-specific code into risk_flags (not just composable_unavailable)
-    # so clients branching on risk_flags alone can tell stale / load_error /
-    # schema_mismatch / invalid_dir apart.
-    risk_flags: list[str] = []
-    if state != DepgraphState.PRESENT:
-        risk_flags.append("composable_unavailable")
-    if blocked_code:
-        risk_flags.append(blocked_code)
+    risk_flags = (
+        ["composable_unavailable", blocked_code]
+        if state != DepgraphState.PRESENT
+        else []
+    )
     model = DepgraphStatusResult(
         state=state,
         gitnexus_dir=status.gitnexus_dir,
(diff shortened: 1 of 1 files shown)


sgathrid (approved):
Nice, clean fix. Approving!

One non-blocking note for a follow-up (no change needed to merge this): the old code guarded the append with `if blocked_code`, and `blocked_by` just above still does. Since `blocked_code` is typed `str | None`, a future non-`PRESENT` state with a `None` code could trip validation when building `risk_flags`. Keeping the guard would future-proof it, for example:

```python
risk_flags = (
    ["composable_unavailable", *([blocked_code] if blocked_code else [])]
    if state != DepgraphState.PRESENT
    else []
)
```

Totally fine to leave as-is for now since every current non-`PRESENT` state has a code. Leaving the **squash-and-merge** to you 🎉

harshjainnn:
I will make these changes and commit them

jeremy-wayland (approved):
Clean! Cheers @harshjainnn, nice first contribution 🎉

harshjainnn:
Thanks @jeremy-wayland glad to contribute

PR #109, opened 2026-07-02
perf(cli): cut startup latency, slim binary onefile, remove semantic coverage 

### LINKED ISSUE #103: Remove Semantic Coverage
### What's the issue?

Not scalable enough. If it's too slow to calibrate locally for the topos leaderboard, its too slow! Every measure should be grounded in real experiments.

### Method / function

topos.functors.profunctors.cpg.topological_coverage

### Details

This was an experimental method for doing ECT comparisons between code snippets and relevant tests. Requires embedding CPGs and then computing ECT and comparing. Currently implementation not scalable enough for the users we are targeting. Let's phase out and see if we can scale this up. Otherwise, maybe ECT can be used rather for other graph comparisons.

### LINKED ISSUE #108: CLI startup is slow for trivial commands (--version, --help)
## Summary

The PyInstaller **onefile** release binary feels bloated for trivial invocations. `topos --version` and `topos --help` were slow because every run paid onefile extraction cost **and** eagerly imported the full evaluation/MCP stack before Click could print anything.

## Command

`topos --version`, `topos --help`

### Observed behavior
- Simple commands took noticeably long compared to other CLIs (e.g. `uv`, `rg`).
- Root cause is twofold:
  1. **PyInstaller onefile extraction** to `_MEIPASS` on every invocation (scales with bundle size).
  2. **Eager Python imports** at startup (`topos/__init__.py`, `commands/quality.py`, etc.) loading graphs, evaluation, and MCP stacks even when no subcommand runs.

### Proposed / implemented approach (branch `feat/cli-startup-optimization`)

**Startup (Python-side)**
- Fast path for root `--version` / `--help` before command registration
- Deferred subcommand registration via `ToposGroup`
- Lazy `topos` package exports (PEP 562)
- Lazy imports inside CLI command handlers
- Lightweight `topos/graphs/ast/languages.py` for CLI option wiring

**Packaging (still onefile)**
- **Default** release artifact: slim `topos-{platform}` without ECT deps (`onnxruntime`, `fastembed`, `trailed`)
- **Optional** artifact: `topos-ect-{platform}` for offline topological coverage
- VS Code extension continues to bundle the slim binary

**Measurement**
- Benchmark harness: `tests/benchmarks/test_cli_startup.py` (`TOPOS_BENCHMARK=1`)
- Docs: `docs/cli-startup-benchmarks.md`
- CI `cli-startup` job: build slim Linux binary + warm startup budgets

### Local results (post-change)

| Channel | `--version` warm | `--help` warm |
|---------|------------------|---------------|
| `uv run topos` | ~59 ms | ~57 ms |
| Slim onefile binary (macOS arm64) | ~1.1 s | ~1.1 s |

Dev path is ~20× faster. Binary warm starts are still dominated by onefile extraction; **onedir** remains a follow-up if we need sub-second binary `--version`.

### Follow-ups (out of scope for initial PR)
- PyInstaller **onedir** packaging for faster warm starts
- Further audit of `--collect-all fastmcp` / `ladybug` for non-MCP commands

### LINKED ISSUE #110: Dogfooding Tests for Binary Install
### What's the issue?

currently no testing to ensure binary works after build

### Workflow / step

after a release --> binary testing

### Details

- start-up time testing?
- `topos --version` testing to ensure `.toml` has properly loaded and versioned?

jeremy-wayland:
Closes #108

## Problem
`topos --version`/`--help` were slow — importing the CLI transitively pulled in evaluation morphisms, graph reps, MCP registrations, and tree-sitter backends before Click parsed argv.

## Fix (3 layers)
1. `cli/main.py` — fast path answers `--version`/`--help` before Click loads; subcommands register lazily.
2. `topos/__init__.py` — only `__version__` loads eagerly; rest lazy via `__getattr__`.
3. `cli/commands/*.py` — heavy imports moved into command bodies.
4. `graphs/ast/dispatch.py` → `languages.py` split — isolates pure data (`SUPPORTED_LANGUAGES` etc.) needed at import time from tree-sitter backends.

## Also includes #103 (merged via #116)
Removes ECT semantic coverage end-to-end (code, CLI/MCP/policy call sites, release binary variant, optional deps). Retires slim-vs-ect packaging split — one binary now.

## Packaging
- Single `topos-{platform}` binary (no more `topos-ect-*`)
- Targeted `--hidden-import`s replace blanket `--collect-all topos`
- **Size:** 72.14 MB → 44.95 MB (-37.7%, this PR) → 39.43 MB (-12.3%, via #116)
- **Warm `--version`:** ~854ms → ~610ms → ~586ms

## Correctness fixes
- `languages.py` split missed 2 importers (`mcp/evaluation.py`, `mcp/tools/evaluate/project.py`) — fixed.
- Installed console script bypassed fast path → empty `Commands:` list on `--help` — fixed by registering in `parse_args()`; regression tests added.
- Closes #110 — CI/release now dogfood the built PyInstaller binaries: `tests/packaging/` smoke tests (version, help, subcommands) run against `TOPOS_BINARY` after each binary build, so a broken frozen binary fails CI instead of shipping.

## Measured impact (macOS arm64)
| Channel | `--version` | `--help` |
|---|---|---|
| `uv run topos` (dev) | ~59ms | ~57ms |
| Binary, pre-PR | ~854ms | ~791ms |
| Binary, post hidden-imports (44.95MB) | ~610ms | ~615ms |
| Binary, post #116 ECT removal (39.43MB) | ~586ms | ~565ms |

## Test plan
- [x] `pytest` / `ruff` clean
- [x] Cold-start benchmark
- [x] Manual: `--help`/bare invocation via both entry points
- [x] Manual: built binary — version/help/evaluate/coverage/mcp all work
- [x] CI `cli-startup` job passes

## Not in this PR
Onedir packaging (~9x faster warm start) — needs installer/VS Code/notarization changes; deferred.


sgathrid (changes requested):
DN --> running speedy when I test. That said, came back with several items to flag. 

## 1. 🔴 Frozen binaries ship without MCP docs → `topos_get_doc` crashes

**Where:** `.github/workflows/release.yml` (`PYINSTALLER_ARGS`), mirrored in `ci.yml`

**Issue:** Dropping `--collect-all topos` stops bundling package **data files**. The replacement `--hidden-import` flags collect Python modules only, and the sole `--add-data` is `Cargo.toml` — nothing re-adds `topos/mcp/resources/content/*.md`.

**Why it matters:** Every frozen binary (slim *and* ect) ships without the six doc topics. `topos mcp` + `topos_get_doc(topic="workflows")` hits a bare `path.read_text()` in `topos/mcp/tools/docs.py:56` (no existence check) → unhandled `FileNotFoundError` for **every** docs topic. This worked in v0.3.7. The `cli-startup` CI job only exercises `--version`/`--help`, so CI never catches it.

**Fix:** Add `--add-data topos/mcp/resources/content:topos/mcp/resources/content` (or `--collect-data topos`) to all binary variants, and extend the CI smoke test to invoke one MCP docs read against the built binary.

---

## 2. 🔴 `languages.py` split is ineffective — subcommands still eagerly load tree-sitter

**Where:** `topos/graphs/ast/__init__.py:7`

**Issue:** `topos.cli.commands.quality`/`coverage` import `topos.graphs.ast.languages` at module top, but importing any submodule first executes the package `__init__`, which does `from topos.graphs.ast.dispatch import ...` → both parser providers → the full tree-sitter stack. Verified empirically: after `_register_commands()`, `sys.modules` contains `dispatch`, `tree_sitter_provider`, and all `tree_sitter_*` bindings.

**Why it matters:** Every subcommand invocation — including `topos update` and `topos mcp`, which never parse code — pays the tree-sitter import at registration time. A broken/missing tree-sitter wheel makes even `topos update` crash with `ImportError`. Only bare `--version`/`--help` (the argv fast path) get the intended isolation, so the split's stated purpose doesn't hold.

**Fix:** Empty out `topos/graphs/ast/__init__.py` (or make its re-exports lazy via PEP 562 `__getattr__`) so `languages` imports stand alone. Bonus: once registration is genuinely cheap, `_ROOT_HELP` and its drift test (`test_root_help_matches_click_generated_help`) can likely be deleted and Click can render `--help` itself. Consider a guard test asserting `tree_sitter`/`topos.mcp` are absent from `sys.modules` after `_register_commands()` so this can't regress silently.

---

## 3. 🟠 `topos update` silently downgrades an ect install to slim

**Where:** `topos/cli/update.py` (`_run_binary_update`), `install.sh:277`

**Issue:** The update flow re-runs `install.sh`, which hardcodes `asset_name="topos-${platform}"`. Nothing detects which variant is installed, and this PR introduces the ect variant.

**Why it matters:** A user running `topos-ect-<platform>` who accepts the update nag gets replaced with the slim binary; their next `topos coverage --topological` silently degrades to the "pip install 'topos-mcp[ect-coverage]'" hint. (Mitigating: ect installs are currently manual placements, not install.sh-managed — but the update notice still fires on them.)

**Fix:** Record the variant in the install provenance receipt (or detect ECT availability at update time) and pass e.g. `TOPOS_VARIANT=ect` through to `install.sh`; at minimum, warn instead of silently swapping.

---

## 4. 🟠 PyInstaller args hand-duplicated in 3 places; hidden-imports mirror `_LAZY_EXPORTS` with no parity check

**Where:** `.github/workflows/ci.yml` (~line 83), `.github/workflows/release.yml`, `extensions/vscode/workflow/publishing.md`

**Issue:** The ~35-line slim invocation exists in three hand-synced copies, guarded only by a "Keep in sync" comment. The 11-entry `--hidden-import` list also hand-mirrors `_LAZY_EXPORTS` in `topos/__init__.py` (in sync today, no test enforcing it).

**Why it matters:** A flag added in `release.yml` but not `ci.yml` means the startup-budget gate measures a binary that isn't what ships — defeating the job's stated purpose. A new `_LAZY_EXPORTS` entry without workflow updates ships a binary that raises `ModuleNotFoundError` on first attribute access. Finding 1 is an instance of exactly this bug class.

**Fix:** Extract to a shared `scripts/build-binary.sh` (or checked-in `.spec`) taking a `slim|ect` flag, invoked by both workflows and referenced from the docs. For hidden imports, a repo-local PyInstaller hook deriving `hiddenimports` from `topos._LAZY_EXPORTS` makes the code the source of truth.

---

## 5. 🟡 "Cold" benchmark isn't cold, and the `_MEI` sweep can break other running apps

**Where:** `tests/benchmarks/test_cli_startup.py:90`

**Issue:** (a) The cold test clears `_MEI*` dirs **once**, then asserts the *median* of 5 runs — runs 2–5 are warm, so the median is a warm figure and `TOPOS_VERSION_COLD_BUDGET_S` never gates actual cold start. (b) The sweep `rmtree`s **every** `_MEI*` dir in TMPDIR with no ownership check, including live extraction dirs of unrelated running PyInstaller apps (env-gated behind `TOPOS_BENCHMARK=1`, but still a dev-machine hazard).

**Fix:** Clear `_MEI*` before *each* timed run and assert the max (or first) run for the cold test; restrict the sweep to dirs created by this test's binary, or use a dedicated `TMPDIR`.

---

## 6. 🟡 Lazy `__getattr__` breaks `topos.core`/`topos.graphs` attribute access and empties `dir(topos)`

**Where:** `topos/__init__.py:65`

**Issue:** v0.3.7's eager imports bound submodules as attributes; now `import topos; topos.core` raises `AttributeError`, and with no PEP 562 `__dir__`, `dir(topos)` lists none of the 14 public exports (breaks introspection/IDE completion/`help(topos)`).

**Why it matters (low severity):** `from topos import X` still works and the package is consumed mainly as CLI/MCP, but it's a silent public-API narrowing.

**Fix:** Add `def __dir__(): return sorted(__all__)`; optionally resolve `core`/`graphs`/`evaluation` lazily as submodules too.

---

## 7. 🟡 Registration guard uses `if cli.commands:` truthiness instead of a flag

**Where:** `topos/cli/main.py:49`

**Issue:** Any consumer adding a command to the group before first invocation (`cli.add_command(...)` in a test fixture/plugin/embedding) makes `cli.commands` truthy, so `_register_commands()` returns early and the built-ins are never attached — reproduced: `topos evaluate` fails with "No such command 'evaluate'". No in-repo trigger today; latent fragility.

**Fix:** Use a module-level `_commands_registered` boolean.

---

## 8. 🟢 Version resolution still runs eagerly on the fast path

**Where:** `topos/_version.py:52`, imported at `topos/cli/main.py:11`

**Issue:** `__version__ = get_version()` executes at import time, before `_fast_path` runs. In the frozen binary only fastmcp metadata is copied, so `version("topos-mcp")` raises `PackageNotFoundError` on *every* run and falls back to parsing the bundled `Cargo.toml` with tomllib. ~30–50ms of the ~610ms frozen startup.

**Fix:** `--copy-metadata topos-mcp` in the build args, or bake a static `_version_static.py` at build time.

---

### Verified clean

- `dispatch.py` → `languages.py` split: all importers updated, no stale references (the `mcp/evaluation.py` regression noted in the PR description is fixed at head).
- `_LAZY_EXPORTS` preserves all 14 previously exported names; hidden-import lists match it exactly today.
- Fast path is behavior-equivalent to Click for `--version`/`--help`; `_ROOT_HELP` drift is pinned by test (note: it is load-bearing until finding 2 is fixed).
- MCP tool registration is fully static, so removing `--collect-all topos` is safe for *modules* — the gap is data files only (finding 1).

jeremy-wayland:
@sgathrid — thanks for the review, addressed all 8 findings on `feat/cli-startup-optimization` (12 commits on top of #116):

| # | Finding | Fix | Commit(s) |
|---|---|---|---|
| 1 🔴 | Frozen binaries ship without MCP docs | Centralized args in `build-binary.sh` (`--add-data .../resources/content`); added `test_frozen_mcp_docs.py` + CI smoke test | `31b7bfd`, `70868a0`, `43fc1b4` |
| 2 🔴 | `languages.py` split ineffective — tree-sitter still eager | PEP 562 lazy exports in `graphs/ast/__init__.py` & `utils/__init__.py`; deferred `parse_source` import in `morphism.py`; guard test asserts no `tree_sitter*` in `sys.modules` | `42c808c`, `a7368e0`, `19f8ea4`, `dd673f3`, `37312f5` |
| 3 🟠 | `topos update` downgrades ect→slim | N/A — #116 removed the ect variant entirely | — |
| 4 🟠 | PyInstaller args duplicated across 3 files | `build-binary.sh` is now single source of truth; `lazy_exports.py` + parity test vs `_LAZY_EXPORTS` | `31b7bfd`, `70868a0`, `95bc4fe` |
| 5 🟡 | "Cold" benchmark isn't cold; `_MEI` sweep unsafe | Clear `_MEI*` before each run, assert max latency, dedicated `TMPDIR` | `fc4f731` |
| 6 🟡 | Lazy `__getattr__` breaks `dir(topos)`/`topos.core` | Added `__dir__()`; lazy-resolve `core`/`graphs`/`evaluation` | `49c82ed` |
| 7 🟡 | Registration guard uses truthiness | Module-level `_commands_registered` flag | `af2406a` |
| 8 🟢 | Version resolution runs eagerly | `--copy-metadata topos-mcp` in build script | `31b7bfd` |

**Verification:** `pytest` — 444 passed, 5 skipped; `ruff check`/`format --check` clean.

sgathrid (approved):
> [!WARNING]
> Blocked by PR #119 

Everything else looks great! Approving for merge here, **_pending_** blocking PR review + acceptance

sgathrid (approved):
Good to go!

PR #120, opened 2026-07-06
MCP: ranked targets + gate/security unification

sgathrid:
## Summary

Evaluation results told agents *what score* they got, not *what to edit*. This PR closes the loop — and fixes some additional design gaps that surfaced in the process (parallel gate encodings, contract contradictions, a depgraph freshness blind spot).

**One call, everything an agent needs:**

```
topos_evaluate_file(filepath="mod.py", refactor_targets=5)
  → ranked edit targets (span + metric + threshold + operations)
  → self-consistent agent contract (next_tool, next_actions, blocked_by)
```

## What changed

| Area | Before | After |
|---|---|---|
| **Edit targets** | Passive scores; agent deduces what to edit | `refactor_targets: int = 0` returns up to N ranked targets with failing metric, current vs. threshold, and `recommended_operations` |
| **Gate logic** | Re-encoded in 4 places (scorers, suggestions, interpretations, target specs); suggestions missed the entrypoint carve-outs and fired on gates the scorer passed | One canonical spec table (`policies/gates.py`) drives all four — divergence is impossible by construction |
| **Security guidance** | Two divergent maps; exact-key lookup missed qualified callees; `subprocess.Popen` got `os.popen`'s advice | One table (`security_guidance.py`), suffix-matched with the danger probe's own matcher; registry coverage is test-enforced |
| **Agent contract** | Post-hoc mutation overwrote `next_tool`/`next_actions` while `blocked_by` said otherwise | Contract builder routes targets natively; invariant: **`next_tool`/`next_actions` never contradict `blocked_by`** |
| **Depgraph generation** | Always regenerated (expensive, side-effecting) | Ensure-style: no-op when current, regenerate when missing/stale, `force=true` to override |
| **Graph freshness** | HEAD-SHA only — in-place edits left a silently stale graph, COMPOSABLE scored the pre-edit tree | Fingerprint v2 `{head_sha, generated_at}` + source-mtime pass catches uncommitted edits; the ensure default then regenerates mid-loop |
| **`SCHEMA_MISMATCH`** | Status routed to regeneration — which can't fix a store newer than the reader | `next_tool = None` + upgrade-Topos / downgrade-GitNexus guidance, consistent across both tools |
| **`include_security_findings`** | Redacted at the overlay → hiding findings also killed security targets, secure suggestions, and the risk flag (secure-first agents misrouted) | Payload gate only: routing always sees true findings; assess/project risk flags anchor on the adjusted verdict, not the redactable list |

## Agent-facing API

| Surface | Change |
|---|---|
| `topos_evaluate_file` | 3 draft params collapsed to `refactor_targets: int = 0` (0 = off, ≤ 25) |
| `RefactorTarget` | `verify_with` removed — verification guidance lives once on `agent_contract.verification_gates`, not repeated per target |
| Discoverability | Below-IDEAL results without targets hint `refactor_targets=5` in `next_actions` |
| Stale cadence | Advisory ("refresh before trusting COMPOSABLE"), so agents batch one regenerate per assess checkpoint instead of per edit |
| Wire budget | Tool surface **33,736 chars** with a deliberate ~750-char ratchet headroom (was 48) |

## Verification

- **632 tests pass**; ruff clean
- `tests/evaluation/test_gate_parity.py` — 89-case characterization grid captured from the pre-refactor scorers; the gate-spec unification reproduces every score, verdict, and interpretation string exactly
- `tests/mcp/test_contract_invariant.py` — {missing, stale, invalid} × targets: blocker survives, remedy rides along
- `tests/evaluation/test_security_guidance.py` — every `DANGEROUS_APIS` entry resolves to specific guidance
- End-to-end: uncommitted edit → status reports STALE (naming the file) → ensure regenerates → secure-first ranking surfaces the `eval()` target first even with findings hidden


jeremy-wayland:
Reviewed this PR (8-angle finder pass + verification). Most findings turned out to be by-design (`include_security_findings` intentionally not redacting `refactor_targets`/`suggestions`, the 20k-file walk cap tradeoff) or pre-existing (the `match_registry_key` suffix-match false-positive risk predates this PR). Three were genuine gaps in the new freshness check, so I opened #122 into this branch with fixes + tests:

- `_newer_source_file` missed deleted/renamed source files (no mtime of its own to catch).
- `mtime > generated_at` had no tolerance for coarse filesystem resolution / clock skew.
- `topos_generate_depgraph`'s SCHEMA_MISMATCH message duplicated `_STATE_GUIDANCE` instead of reusing it.

@sgathrid — flagging for your review since #122 targets this branch.

sgathrid:
@jeremy-wayland merged in your changes, tests working now.

Awaiting your final review/approval!

krv-labs-ai-developer:
Resolved merge conflicts and updated CHANGELOG.md with PR120 changes (Go language support, CFG/CLI fixes).

jeremy-wayland:
Small hardening pass on top of @sgathrid's freshness work (`60b8820`):

- **Duration clamp**: the mtime-drift calibration in `_newer_source_file` now sanity-bounds the measured `finished_at - generated_at` duration before trusting it, falling back to the existing flat tolerance if it's negative or implausibly large (e.g. a mid-generation clock adjustment) instead of extrapolating a threshold from a bad sample.
- **Debug-level observability**: tags which freshness method (`content_hash` / `sha_anchor` / `mtime_calibrated` / `sha_only_no_signal` / `legacy_dir_mtime`) decided each verdict, so a suspicious result is diagnosable from logs rather than only from reading the source. Debug-only — doesn't touch the public `DepgraphStatus` response.

New regression test (`test_freshness_v2_distrusts_implausible_duration`) confirms a real edit is still caught under the clamp — verified it fails without the fix. All existing freshness tests pass unmodified.

jeremy-wayland (approved):
Great work @sgathrid 🏋️

PR #122, opened 2026-07-07
fix(mcp): close depgraph freshness gaps found in review of #120

jeremy-wayland:
Follow-up from reviewing #120. Three of the findings there were genuine gaps in the new fingerprint-v2 freshness check (not by-design behavior); this fixes those and leaves the rest as-is (see rationale below).

## Fixed here

- **Deleted/renamed source files were invisible to `_newer_source_file`.** It only walked files that still exist, so removing or renaming a file after generation (HEAD unchanged) left the graph reporting "fresh" forever. Directory mtimes are now checked too, after that directory's own file children (so a plain edit still names the exact file — only a deletion/rename falls back to naming the containing directory).
- **No tolerance in the `mtime > generated_at` comparison.** `generated_at` is a sub-second `time.time()` stamp compared against filesystem mtimes, which can be coarser or clock-skewed. Added a small grace window so an edit made moments after generation doesn't silently read as fresh.
- **`topos_generate_depgraph`'s SCHEMA_MISMATCH branch hand-duplicated the guidance text** that already lives in `_STATE_GUIDANCE` (used by `topos_depgraph_status`). Now sources both the message and `next_actions` from the same table, matching the PR's own "single source of truth" thesis.

`iter_source_files` gained an `include_dirs` opt-in param (default off, no behavior change for existing callers) so the freshness check gets directory signals from the same walk instead of a second pass.

## Deliberately left alone

- The `_FRESHNESS_WALK_CAP` silent-fresh-past-20k-files behavior — already documented as an intentional perf/correctness tradeoff (SHA anchor still catches commit-level drift).
- `include_security_findings` not redacting `refactor_targets`/`suggestions` evidence — that's the PR's explicit, documented design ("routing and guidance always derive from true findings"), not a bug.
- The bare-suffix false-positive risk in `match_registry_key` (e.g. `safe_eval` matching `eval`) — pre-existing logic carried over from `_matches_registry`, not introduced by #120. Worth its own issue, but out of scope for this branch.

## Verification

- Added `test_freshness_v2_deleted_file_is_stale` and `test_freshness_v2_tolerates_small_mtime_skew` to `tests/mcp/test_depgraph.py`.
- Full suite: 632 passed, 10 skipped (630 pre-existing + 2 new).

### DIFF
--- a/tests/mcp/test_depgraph.py
+++ b/tests/mcp/test_depgraph.py
@@ -497,18 +497,70 @@ def test_freshness_v2_deleted_file_is_stale(tmp_path) -> None:
 
 
 def test_freshness_v2_tolerates_small_mtime_skew(tmp_path) -> None:
-    # generated_at is a sub-second time.time() stamp; filesystem mtimes can
-    # be coarser or drift slightly against the process clock. An edit inside
-    # the tolerance window must still read as stale, not silently as fresh.
+    # Simulate a system with clock drift (e.g., filesystem clock is 5.0 seconds behind).
+    # We edit `f.py` after the graph was generated, but due to clock drift,
+    # the filesystem writes `f.py` with an mtime that appears earlier than the process's `generated_at`.
+    # With dynamic drift compensation, the drift is calculated using the fingerprint file's own mtime
+    # on the same filesystem, and the edit is still correctly detected as stale.
+    import os
     head = _commit_repo(tmp_path)
     gitnexus = _graph_dir(tmp_path, fingerprint=None)
-    source_mtime = (tmp_path / "f.py").stat().st_mtime
-    _write_v2_fingerprint(gitnexus, head_sha=head, generated_at=source_mtime + 1.0)
+    
+    # Write fingerprint file with generated_at=100.0, finished_at=100.0
+    (gitnexus / GITNEXUS_FINGERPRINT_FILE).write_text(
+        json.dumps({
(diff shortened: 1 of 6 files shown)


sgathrid (changes requested):
Not sure the point at which we just are pitting `/review` agents against each other, but got this re your changes:

> 
> found one remaining blocker in the freshness/ensure contract.
> 
>  [P2] Don’t mark freshly generated graphs stale due to the mtime tolerance
> 
>   `topos/mcp/evaluation.py:215`
> 
>   `_newer_source_file()` compares source and directory mtimes against `generated_at -2s`. That catches edits made shortly after generation, but it also treats files modified shortly before generation as stale. In the ensure path, a graph generated immediately after an edit can therefore be reported stale even though GitNexus indexed that edit, so `topos_generate_depgraph` may keep regenerating instead of no-oping when the graph is actually current. 
>
>   The branch’s own `test_freshness_v2_tolerates_small_mtime_skew` encodes this: it writes `generated_at = source_mtime + 1.0` and expects stale. For the ensure- style contract, that ordering is also the normal “source existed before the graph was generated” case. We still need to catch uncommitted edits/deletions after generation, but the tolerance needs to avoid classifying pre-generation mtimes inside the skew window as stale/current-loop work.

sgathrid (approved):
gtg!

PR #126, opened 2026-07-08
fix: remove #[cfg(test)] nodes from AST build

krv-labs-ai-developer:
Filtering out rust attributes marked as #[cfg(test)] during AST generation to exclude unit tests from the structural analysis, as per Jeremy's feedback.


krv-labs-ai-developer:
Hi @jeremy-wayland,

I've implemented the scoped fix for filtering  modules in the Rust mapper.

**Suggestion:**
This type of test filtering is highly relevant to other languages as well (e.g., Python's  functions or JS / blocks). Instead of keeping language-specific filters in , I suggest we move this toward a generalized 'test detection' interface where each language mapper defines what constitutes a 'test' node.

Would you be open to helping me design a cleaner, language-agnostic interface for this? I'd love your input on the best way to keep the common mapper clean while supporting these language-specific quirks.

jeremy-wayland (approved):
👍

PR #131, opened 2026-07-11
perf(cli): avoid eager numpy/eval import in quality.py registration (#119)

### LINKED ISSUE #119: perf(cli): command registration eagerly loads numpy/eval stack via quality.py Priority import
Follow-up carved out of the #109 review (see #118). Deliberately **not** fixed there — this issue records the analysis so the priority is clear before anyone picks it up.

## What

`_register_commands()` imports `topos/cli/commands/quality.py`, whose module-level `from topos.evaluation.policies.base import Priority` transitively pulls `preferences → characteristic_morphism → functors.profunctors.ast.compare → numpy`. So every real subcommand invocation loads the full evaluation stack + numpy at registration, even commands that never evaluate code.

`system.py` is already clean — `quality.py` is the sole trigger. Confirmed empirically:
- `import topos.cli.commands.system` → numpy **not** loaded
- `import topos.cli.commands.quality` → numpy loaded

## Not a regression

This is **pre-existing** and predates #109. Before #109, `quality.py` was imported eagerly at `main.py` load and pulled numpy + `characteristic_morphism` + tree-sitter on *every* invocation, including `topos --version`. #109 fixed the review's actual finding (tree-sitter no longer loads at registration) and added the `--version`/`--help` fast path. This numpy residual is what's left.

## Measured gain (small, narrow)

| | Cost |
|---|---|
| `_register_commands()` total | ~185 ms (dev interpreter) |
| numpy/eval slice of that | **~45–78 ms** (dev); larger in the frozen onefile binary |

Beneficiaries are only:
- `topos update`, `topos uninstall` (~0.22 s → ~0.15 s, dev path)

Everyone else gets **zero**:
- `--version` / `--help` — already fast-pathed, never reach registration
- `evaluate` / `coverage` / `inspect` / `compare` / `mcp` — genuinely need the eval stack; loading it is not waste

So: a small win on two infrequent, non-interactive commands. **Low priority** unless frozen-binary `topos update` latency becomes a concern.

## Why it isn't trivially fixed (pre-mortem)

- **Relocate the `Priority` enum to a numpy-free module** — looks clean, but `Priority` is **not** vestigial: it's a live type across ~25 files (core `characteristic_morphism`, every policy scorer, MCP Pydantic field types, `assess/snapshot.py`'s `Priority(meta[...])`). High blast radius, touches serialized model fields. Not worth it.
- **Make the `policies` package lazy (PEP 562)** — foundational, widely-imported package; risks star-imports / pickling / `dir()`. Highest risk.
- **Shared trap:** the "cheap registration" invariant is currently guarded only for `tree_sitter` (`tests/cli/test_registration_imports.py`); nothing pins numpy/eval absence, so any fix can silently regress. A numpy guard can't be added *today* because numpy loads now — it must land *with* the fix.

## Recommended approach if picked up

Scoped, ~15 lines, fully guarded — do **not** relocate `Priority`:

1. In `quality.py`, build `_PRIORITY_CHOICE` and the `--priority` default from a module-level literal (`("simple", "composable", "secure")`) so no eval import happens at module scope. Use the `Priority` enum only inside command callbacks (which already lazy-import their heavy deps). Note: quality.py only ever needs the `.value` strings, not the enum type, at module scope.
2. Add a parity test: `set(_PRIORITY_CHOICE.choices) == {p.value for p in Priority}` so the literal can't drift from the enum.
3. Extend `tests/cli/test_registration_imports.py` to assert `numpy` and `topos.evaluation` stay out of `sys.modules` after `_register_commands()`, locking the invariant in.

## Acceptance

- `_register_commands()` no longer loads `numpy` / `topos.evaluation.*`
- `--priority` still accepts exactly the `Priority` values, guarded by a parity test
- Registration-import guard test covers numpy/eval, not just tree-sitter

jeremy-wayland:
## Summary

Closes #119.

`_register_commands()` imported `topos/cli/commands/quality.py`, whose module-level `from topos.evaluation.policies.base import Priority` transitively pulled in the full evaluation stack (`characteristic_morphism`, `functors.profunctors.ast.compare`) and `numpy`, on every real subcommand invocation — even commands that never evaluate code.

Implements the recommended fix from the issue (not the two rejected alternatives — relocating `Priority` or making `policies` lazy via PEP 562):

- `topos/cli/commands/quality.py`: replaced the module-scope `Priority` import with a module-level literal `_PRIORITY_VALUES = ("simple", "composable", "secure")`, used to build `_PRIORITY_CHOICE` and the `--priority` option defaults (`"secure"`). `Priority` is no longer imported anywhere in this file — the `evaluate`/`inspect` callbacks already lazy-import their heavy deps (`run_classify_file`, which internally does `Priority(priority)`), and the module-scope preference-validation checks now compare against `_PRIORITY_VALUES` directly.
- `tests/cli/test_quality.py`: added `test_priority_choice_matches_priority_enum`, a parity test asserting `_PRIORITY_VALUES`/`_PRIORITY_CHOICE.choices` match `{p.value for p in Priority}`, so the literal can't silently drift from the enum.
- `tests/cli/test_registration_imports.py`: added `test_register_commands_does_not_import_numpy_or_evaluation`, extending the existing tree-sitter import guard to also assert `numpy` and `topos.evaluation.*` stay out of `sys.modules` after `_register_commands()`.

## Acceptance criteria

- [x] `_register_commands()` no longer loads `numpy` / `topos.evaluation.*` — verified empirically and via the new guard test.
- [x] `--priority` still accepts exactly the `Priority` values, guarded by a parity test.
- [x] Registration-import guard test covers numpy/eval, not just tree-sitter.

## Test plan

- [x] `uv run pytest --no-cov -q` — 465 passed, 10 skipped, no failures/regressions.
- [x] `uv run pytest --no-cov -q tests/cli/test_registration_imports.py tests/cli/test_quality.py` — 22 passed.
- [x] `uv run ruff check --fix` / `uv run ruff format` on changed files — clean, no changes needed.
- [x] Manually confirmed `_register_commands()` leaves `numpy` / `topos.evaluation.*` out of `sys.modules`.

No deviations from the plan in the issue.

### DIFF
--- a/tests/cli/test_quality.py
+++ b/tests/cli/test_quality.py
@@ -344,3 +344,14 @@ def test_evaluate_verbose_lists_findings(tmp_path: Path):
     out = _strip_ansi(result.output)
     assert "Security Findings" in out
     assert "eval" in out
+
+
+def test_priority_choice_matches_priority_enum():
+    """quality.py's module-level ``_PRIORITY_VALUES`` literal (kept import-free
+    to avoid eagerly loading the eval stack / numpy at CLI-registration time)
+    must not drift from ``Priority``, the enum it mirrors."""
+    from topos.cli.commands.quality import _PRIORITY_CHOICE, _PRIORITY_VALUES
+    from topos.evaluation.policies.base import Priority
+
+    assert set(_PRIORITY_VALUES) == {p.value for p in Priority}
+    assert set(_PRIORITY_CHOICE.choices) == {p.value for p in Priority}
--- a/tests/cli/test_registration_imports.py
+++ b/tests/cli/test_registration_imports.py
@@ -27,3 +27,30 @@ assert not tree_sitter, f"unexpected tree_sitter imports: {tree_sitter}"
         text=True,
     )
     assert completed.returncode == 0, completed.stdout + completed.stderr
+
+
+def test_register_commands_does_not_import_numpy_or_evaluation() -> None:
+    """Registering commands (incl. quality.py's ``--priority`` option) must
+    not eagerly pull in numpy or the evaluation stack. Those are heavy and
+    only needed once a command actually evaluates code, inside lazy-imported
+    callbacks (see issue #119)."""
+    script = """
+import sys
+
+from topos.cli.main import _register_commands
+
+_register_commands()
+
+heavy = [
+    name for name in sys.modules
+    if name == "numpy" or name.startswith("numpy.")
+    or name == "topos.evaluation" or name.startswith("topos.evaluation.")
+]
+assert not heavy, f"unexpected eager imports: {heavy}"
+"""
+    completed = subprocess.run(
+        [sys.executable, "-c", script],
+        capture_output=True,
+        text=True,
+    )
+    assert completed.returncode == 0, completed.stdout + completed.stderr
--- a/topos/cli/commands/quality.py
+++ b/topos/cli/commands/quality.py
@@ -5,11 +5,16 @@ from pathlib import Path
 
 import click
 
-from topos.evaluation.policies.base import Priority
 from topos.graphs.ast.languages import SUPPORTED_LANGUAGES, language_file_suffixes
 
 _EVALUATE_LANGUAGE_CHOICE = click.Choice(sorted(SUPPORTED_LANGUAGES))
-_PRIORITY_CHOICE = click.Choice([p.value for p in Priority])
+
+# Mirrors topos.evaluation.policies.base.Priority's values without importing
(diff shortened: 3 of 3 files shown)


krv-labs-ai-developer (approved):
I have evaluated this PR and verified the fix for issue #119.

The implementation correctly decouples the CLI registration from the heavy evaluation stack imports by using a literal tuple for priority values. I have verified the fix with the new registration guard tests and confirmed that they pass in the local test environment.

The solution is clean, robust, and correctly addresses the performance bottleneck while including parity tests to prevent future drift.

PR #133, opened 2026-07-11
refactor(uast): language-agnostic test-node filtering interface (#127)

### LINKED ISSUE #127: Generalize AST test/source code separation across languages
## Description
We need a generalized, language-agnostic interface to distinguish between test-only nodes and source code structure during AST generation (e.g.,  in Rust,  functions in Python, / blocks in JS/TS).

Currently, we are patching language-specific filters into , which is not scalable.

## Proposed Solution
Introduce a language-agnostic interface for identifying 'test' nodes, allowing each language mapper to provide its own predicates without cluttering the core transformation engine.

## Related
- PR #126: Scoped fix for Rust.
- Issue #124: Feedback on COMPOSABLE pillar and Rust architectural metrics.

jeremy-wayland:
## Summary

Closes #127. Replaces the hardcoded `if language != "rust": return named` branch inside `mapper_common.py`'s `_filtered_named_children` (added by #126) with a language-agnostic predicate interface, so per-language "is this a test-only node" logic lives in each language's mapper instead of the shared transformation engine.

## Interface shape

```python
# topos/graphs/uast/mapper_common.py
TestNodePredicate = Callable[[Node, list[Node]], bool]
```

Given a candidate node and the *full ordered list of its named siblings* (including itself), the predicate returns `True` if that node — and its whole subtree — should be excluded from the UAST.

Why siblings, not just the node: Rust's `#[cfg(test)]` marker lives on a **separate preceding sibling** node, not inside the item it annotates, so a single-node predicate can't express "drop the item after a `#[cfg(test)]` attribute" without seeing adjacent siblings. Python's `if __name__ == "__main__":` guard is fully self-contained, so its predicate just ignores the second argument (`del siblings`).

`is_test_node` is threaded through `map_tree_sitter_to_uast(...)` as an optional keyword parameter, exactly the way `map_node_kind` already is — a per-language callable, not an if/elif chain in the shared function. It defaults to `None`, which preserves today's "map every named node" behavior for any language that doesn't supply one.

I considered a heavier Protocol/dataclass form but went with a plain typed callable (`TestNodePredicate = Callable[...]`) per the task's guidance to prefer the lighter option when the shape is ambiguous — there's no per-predicate state or configuration needed beyond "given this context, yes or no."

## Which languages got what

| Language | Predicate | Behavior |
|---|---|---|
| Rust (`mapper_rust.py`) | `is_test_node` — migrated | Same `#[cfg(test)]` sibling-correlation logic as before #127, moved out of `mapper_common.py` verbatim (same forward-scan state machine, just evaluated per-candidate against the sibling list so it fits the new callable shape). Existing `test_rust_ast_filtering.py` passes **unmodified**. |
| Python (`mapper_python.py`) | `is_test_node` — new | Drops `if_statement` nodes whose `condition` field is a `comparison_operator` matching `__name__ == "__main__"` (either operand order), taking the guard's whole body with it. New test file `tests/graphs/ast/test_python_ast_filtering.py` mirrors the Rust test's structure (asserts on surviving *native node kinds*, not span-text substring checks, to guard against the same class of bug #126 fixed) and additionally covers reversed-operand order and confirms unrelated `if` statements that merely reference `__name__` or `"__main__"` individually are **not** dropped. |
| JS / TS / C++ / Go | none — pass-through | Untouched. `is_test_node` defaults to `None` in all four mappers, so behavior is bit-for-bit identical to before this PR. JS/TS `describe`/`it` filtering is a natural follow-up once this interface has proven out on Rust + Python, but it needs call-expression heuristics that are a different shape of problem — deliberately out of scope here. |

## Scope guardrail

Per the task, `mapper_javascript.py`, `mapper_typescript.py`, and `mapper_cpp.py` were **not touched** — they call `map_tree_sitter_to_uast` without the new `is_test_node` argument, so they get the `None` default (no filtering), identical to pre-PR behavior. `mapper_go.py` likewise untouched, same reasoning (it wasn't named in the issue but follows the same "no predicate supplied" default).

## Test plan

- [x] `uv run pytest --no-cov -q` — 471 passed, 10 skipped (pre-existing skips, unrelated to this change)
- [x] `tests/graphs/ast/test_rust_ast_filtering.py` passes unmodified — confirms Rust behavior preserved exactly
- [x] New `tests/graphs/ast/test_python_ast_filtering.py` passes — 4 tests covering: guard filtered, `main()` function survives, unrelated `if` statements survive, reversed operand order still filtered
- [x] `uv run ruff check` / `uv run ruff format` clean on all changed files


jeremy-wayland:

PR #155, opened 2026-07-13
fix(cfg): count match/case arms toward cfg.cyclomatic for Rust and Python

### LINKED ISSUE #151: SIMPLE/Rust: cfg.cyclomatic doesn't count match-arm branches
Found while dogfooding v0.3.10's own `topos evaluate --language rust` against new Rust code (v0.4.0 migration worktree, issue #104 spirit).

An 8-arm `match` reports `cfg.cyclomatic: 1.000` — identical to a function with no branching at all:

```rust
    match x {
        0 => "a", 1 => "b", 2 => "c", 3 => "d",
        4 => "e", 5 => "f", 6 => "g", _ => "h",
    }
}
```

`topos inspect` on this file: `cfg.cyclomatic: 1.000 (within threshold)`. Standard cyclomatic complexity counts each decision point +1, so an 8-arm match should read ~8, not 1 — `if`/`for`/`while` are evidently counted by the Rust CFG builder (confirmed separately: replacing several 8-arm matches in a real file with equivalent logic did shift other metrics but left `cfg.cyclomatic` bit-for-bit identical), but `match_arm` apparently isn't.

Likely in the Rust CFG builder (`topos/graphs/cfg/builder.py` and/or the Rust-side `cfg.rs` betti/cyclomatic probe) — whatever walks the UAST/CFG to count branch nodes for Rust doesn't include `match_expression`/`match_arm` in its branch-node kind list. Python's `if`/`elif` and Go's equivalent presumably are covered given #72-74 (Go SIMPLE support); worth checking whether Go's own `switch` has the same gap.

Impact: real Rust code that uses `match` heavily (idiomatic Rust) can systematically under-report SIMPLE complexity, i.e. false negatives on genuinely branchy Rust code. Relevant now that v0.4.0's own Rust rewrite will be evaluated by topos on itself.

jeremy-wayland:
## Summary
- `_match_arms()` in `topos/graphs/cfg/builder.py` only recognized Go's native case-arm node kinds (`expression_case`, `type_case`, `default_case`, `communication_case`), so Rust's `match_block`/`match_arm` and Python's `case_clause` were never detected as arm boundaries.
- Every Rust `match` / Python `match`/`case` silently collapsed to a single `UNCONDITIONAL` edge regardless of arm count, so `cfg.cyclomatic` for an 8-arm match read `1.0` — identical to straight-line code (false negative on genuinely branchy code).
- Fix adds a dedicated `_MATCH_ARM_NATIVE_KINDS` set, kept separate from the existing Go-oriented allowlists to avoid a flattening trap: naively reusing `_CASE_ARM_NATIVE_KINDS` would recurse into each arm's body and merge multi-statement arms into one flat list, losing arm boundaries.
- Go's own switch/select handling is untouched.

## Test plan
- [x] New `tests/graphs/cfg/test_rust_cfg.py`: 8-arm match reproduces the issue directly (`cfg.cyclomatic == 8.0`, 8 `SWITCH_CASE` edges), a multi-statement-arm regression guard, and a nested-if-inside-arm case.
- [x] New `tests/graphs/cfg/test_python_match_cfg.py`: same coverage for Python `match`/`case` (same latent bug, found during investigation, not in the original report).
- [x] Full `pytest` (677 passed, 10 skipped) and `ruff check --fix` / `ruff format` clean.
- [x] Manually reproduced the exact issue snippet before/after the fix.

Fixes #151

### DIFF
--- a/tests/graphs/cfg/test_python_match_cfg.py
+++ b/tests/graphs/cfg/test_python_match_cfg.py
@@ -1,0 +1,89 @@
+"""
+Python-specific `match`/`case` CFG shape regression tests.
+
+Python's tree-sitter grammar wraps `match` arms similarly to Rust:
+`match_statement` -> [subject, block], and that `block`'s children are
+`case_clause` nodes.  Prior to this fix, `_match_arms()` didn't recognize
+`case_clause` as an arm boundary, so every Python `match` silently
+collapsed to a single unconditional edge regardless of case count — the
+same class of bug reported for Rust in issue #151, confirmed to affect
+Python too during investigation.
+
+These tests pin down the corrected CFG shape so future grammar-mapping or
+builder changes don't silently regress it.
+"""
+
+from __future__ import annotations
+
+from collections import Counter
+
+from topos.core.morphism import ProgramMorphism
+from topos.graphs.cfg.models import EdgeKind
+from topos.graphs.cfg.object import ControlFlowGraph
+
+
+def _cfg(source: str) -> ControlFlowGraph:
+    morphism = ProgramMorphism(source=source, language="python")
+    cfg = morphism.build_cfg()
+    assert cfg is not None
+    return cfg
+
+
+def test_python_match_creates_one_arm_per_case():
+    src = (
+        "def classify(n):\n"
+        "    match n:\n"
+        "        case 0:\n"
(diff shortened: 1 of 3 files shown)


jeremy-wayland:

PR #156, opened 2026-07-13
fix(entropy): blend ast.entropy toward neutral baseline below a size floor

### LINKED ISSUE #152: SIMPLE: ast.entropy (gzip-ratio proxy) penalizes small, dense-but-simple functions
Found while dogfooding v0.3.10 during the v0.4.0 Rust migration (issue #104 spirit), as a side-effect of fixing the match-arm cyclomatic gap (see the sibling issue filed alongside this one).

Refactoring an 8-arm lookup `match` (cyclomatic-cheap per the *current* metric, but repetitive) into a 3-line const array + index:

```rust
    const T: [&str; 8] = ["a", "b", "c", "d", "e", "f", "g", "h"];
    T[x as usize]
}
```

trips `ast.entropy: 0.880` ("too high; code may be unstructured", SIMPLE fails), versus `0.549` ("within structured range") for the equivalent match. This is clearly the simpler of the two implementations (one branch-free array index vs. an 8-arm match), but `ast.entropy` — a gzip-compression-ratio proxy for Kolmogorov complexity — scores it worse.

Likely cause: gzip has fixed per-stream overhead that dominates the ratio on very short inputs, and a literal array of 8 distinct short strings compresses less well than 8 repeated `match`/`=>` tokens (which gzip's LZ77 window catches easily). So the entropy proxy ends up measuring "textual repetitiveness of the match keyword," not "structural complexity."

Impact: real refactors that replace repetitive branchy logic with data-driven lookups (a textbook SIMPLE-generator improvement) can get flagged as regressions by the entropy metric, specifically for small functions. Worth checking whether there's a length-based floor/normalization already in `topos/functors/probes/ast/entropy.py` (`calculate_entropy_variance`/block entropy) that should be kicking in here but isn't, or whether the ratio needs a minimum-size guard.

jeremy-wayland:
## Summary
- `calculate_kolmogorov_proxy` (`src/probes_ast.rs`) computed a raw `zlib(source).len() / source.len()` ratio with zero size normalization. zlib's fixed per-stream overhead (header + Adler-32 trailer + minimum block cost) dominates the ratio on short inputs.
- Reported symptom: a 3-line array-lookup refactor of an 8-arm match scored *worse* (entropy 0.880, failing the `<=0.8` SIMPLE gate) than the verbose match it replaced (0.549, passing) — a textbook simplifying refactor flagged as a regression.
- Fix: below a 200-byte floor, blend the raw ratio proportionally toward the neutral `entropy_ideal` (0.5) instead of trusting it outright. At/above the floor, behavior is byte-for-byte unchanged (verified with a dedicated test).
- Two existing tests (`test_evaluate_lowest_hanging_fruit`, `test_assess_improvement_has_no_regression_diff`) had fixtures that unintentionally relied on the old tiny-input entropy bug to produce "hard failure" / lattice-tier-jump behavior; reworked both to fail/pass for legitimate structural reasons (repetitive boilerplate size padded above the floor, cyclomatic-complexity-driven near-miss, broadened accepted status set) so they're no longer coupled to this bug.

## Test plan
- [x] New Rust unit tests in `src/probes_ast.rs`: tiny-dense-function regression test (the issue's own array-lookup snippet), and a test confirming the blend is a no-op at/above the size floor.
- [x] New Python regression test in `tests/functors/probes/ast/test_ast_metrics.py` reproducing the issue's exact array-lookup vs. match snippets end-to-end through the SIMPLE gate band.
- [x] `cargo test` (24 passed) and full `pytest` (672 passed, 10 skipped) green; `ruff check --fix` / `ruff format` and `cargo fmt --check` clean.
- [x] Manually reproduced the exact issue snippet before/after the fix via `uv run maturin develop`.

Fixes #152

### DIFF
--- a/src/probes_ast.rs
+++ b/src/probes_ast.rs
@@ -1,5 +1,31 @@
 use pyo3::prelude::*;
 
+const ENTROPY_SIZE_FLOOR_BYTES: usize = 200;
+const ENTROPY_NEUTRAL_RATIO: f64 = 0.5; // mirrors SIMPLE.entropy_ideal
+
+/// Raw zlib compressed/original byte ratio, blended toward a neutral
+/// baseline below `ENTROPY_SIZE_FLOOR_BYTES`.  zlib's fixed per-stream
+/// overhead (header + Adler-32 trailer + minimum block cost) dominates the
+/// ratio on tiny inputs, so below the floor the raw ratio isn't trusted
+/// outright — it's blended proportionally toward `ENTROPY_NEUTRAL_RATIO`
+/// instead of gating a handful of source bytes on compression trivia.
+fn compressed_ratio(source_bytes: &[u8]) -> (usize, f64) {
+    let mut encoder = flate2::write::ZlibEncoder::new(Vec::new(), flate2::Compression::new(9));
+    use std::io::Write;
+    encoder.write_all(source_bytes).unwrap();
+    let compressed = encoder.finish().unwrap();
+    let compressed_len = compressed.len();
+
+    let raw_ratio = compressed_len as f64 / source_bytes.len() as f64;
+    let ratio = if source_bytes.len() >= ENTROPY_SIZE_FLOOR_BYTES {
+        raw_ratio
+    } else {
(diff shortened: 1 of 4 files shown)


jeremy-wayland:

PR #157, opened 2026-07-13
fix(composable): gate on Distance from Main Sequence, not raw instability

### LINKED ISSUE #124: COMPOSABLE pillar: instability band penalizes intentional layering in Rust crates
# Topos COMPOSABLE feedback (for maintainers)

Feedback from evaluating Rust service crates with Topos (CLI v0.3.9) and GitNexus-backed module dependency metrics. Intended for the Topos maintainers — paste into a GitHub issue, discussion, or email.

---

**Subject:** COMPOSABLE pillar feedback — instability band vs layered crate/module graphs (Rust)

Hi Topos team,

I'm using Topos with GitNexus-backed module dependency metrics to evaluate Rust service crates. **SIMPLE** and **SECURE** align well with our intuition; **COMPOSABLE** often flags well-structured crates as failing even when dependency direction is intentional. Sharing this in case it helps calibrate COMPOSABLE for systems languages and layered architectures.

## Context

We run:

```bash
topos evaluate --language rust --recursive \
  --gitnexus-dir .gitnexus \
  --verbose \
  path/to/crate/src
```

With GitNexus enabled, COMPOSABLE appears driven largely by **`mdg.instability`**, with guidance to keep instability in **0.3–0.7** (“rebalance dependencies”). In a typical Rust crate we consistently see:

- **Instability 0.0** on foundation modules (constants, error types, pure domain) and barrel/re-export modules (`mod.rs`, `lib.rs` shells).
- **Instability 1.0** on orchestration modules (loaders, HTTP wiring, bootstrap code that composes many internal modules).
- **Instability ~0.3–0.5** on a small set of “middle” modules (e.g. format conversion, auth helpers) that both import libraries and are imported by higher layers.

That pattern matches **Robert C. Martin’s instability metric** applied to a **deliberately layered DAG** (dependencies point inward). In Rust, extremes at **I ≈ 0** and **I ≈ 1** are often **desired**, not technical debt.

## Why this feels like a false positive for “composability”

In our reading, COMPOSABLE currently rewards **middle-graph packages** and penalizes:

1. **Stable leaves** (I → 0) — modules that should not acquire outbound dependencies.
2. **Unstable orchestrators** (I → 1) — modules whose job is to wire subsystems at the application boundary.

Penalizing both ends pushes toward **artificial coupling** (e.g. adding imports only to move instability into band) which can **reduce** real composability while improving the score.

So the pillar name “COMPOSABLE” reads as “balanced fan-in/fan-out in the MDG,” not “easy to compose/replace subsystems in practice”—which is what we expected from the lattice wording.

## Rust / GitNexus-specific effects

These may inflate COMPOSABLE failures without reflecting bad design:

- **Crate root (`lib.rs`)** — composition root; high efferent coupling by construction.
- **Barrel modules** (`pub use` only) — often parsed as zero-import leaves (I = 0).
- **Binary entry (`main.rs`)** — thin wrapper can score well (I ≈ 0.5) while the library floor remains SIMPLE_SECURE due to orchestrators.
- **Monorepo GitNexus index** — graph scoped to repo root; within-crate `use crate::…` edges dominate, cross-crate boundaries may be underrepresented.

We also occasionally see a **verdict label** (e.g. IDEAL) that disagrees with a **failed COMPOSABLE pillar line** in `--verbose` output; we treat per-pillar scores as authoritative but it adds confusion.

## What worked well

- **SIMPLE** (cyclomatic + entropy band) and **SECURE** (CPG) were actionable and matched review instincts.
- **Verbose MDG fields** (`mdg.instability`, `mdg.coupling`, suggestions) made failures diagnosable once we understood instability semantics.
- Files with genuine **mixed coupling** (used by others and importing stable libs) pass COMPOSABLE predictably—good sanity check.

## Suggestions for maintainers

1. **Document COMPOSABLE as instability-based** in agent contract / metrics docs, including the 0.3–0.7 band and Martin’s Ce/Ca definition. Many users will read “COMPOSABLE” as substitutability/ports, not package instability.

2. **Language- or pattern-aware allowances** for common non-middle modules, e.g.:
   - `main.rs` / binary entrypoints
   - crate root / module declaration files
   - re-export-only barrels
   - `constants` / `paths` / `error` style leaf modules

   Options: exempt from COMPOSABLE floor, separate sub-score, or wider acceptable band for tagged roles.

3. **Role classification from graph shape** (heuristic): if a module is only `pub use` or has zero functions and only re-exports, don’t fail COMPOSABLE on I=0; if a module’s primary consumers are `main`/server/bootstrap, treat high I as expected for an **application layer** rather than a failure.

4. **Floor vs average reporting** — directory floor (lattice meet) heavily punishes orchestrators in otherwise healthy crates. Consider reporting **COMPOSABLE floor among non-exempt modules** or a **recommended target** separate from hard fail for entry/wiring layers.

5. **Rust guidance in docs** — short section: “Layered Rust crates often fail COMPOSABLE until ports/traits create middle coupling; failing on leaves (I=0) and loaders (I=1) alone may be normal.”

6. **Verdict consistency** — ensure lattice element (IDEAL, SIMPLE_SECURE, etc.) always reflects failed pillars in verbose/JSON output.

## Example of “good” architectural fix vs score gaming

**Good (real composability):** introduce a `BlobStore` / `DatasetSource` trait; domain depends on trait; infra implements it → new balanced edges.

**Gaming (score only):** add unused `use crate::…` in leaf modules to raise instability.

We’d like COMPOSABLE to steer toward the former, not accidentally encourage the latter.

## Summary ask

Please consider whether **instability ∈ [0.3, 0.7]** is the right default COMPOSABLE gate for **all** modules in **Rust (and similar layered) codebases**, or whether COMPOSABLE should distinguish **foundation / domain / application / entry** layers and calibrate expectations accordingly.

Happy to share redacted `--verbose` logs or a minimal sample crate if useful.

Thanks for building Topos—the SIMPLE/SECURE pillars and GitNexus integration are already valuable for agent-driven refactors; COMPOSABLE would be even more so with clearer semantics for layered native code.

jeremy-wayland:
## Summary

- COMPOSABLE gated Martin's instability (`I = Ce/(Ca+Ce)`) against a fixed `[0.3, 0.7]` band, flagging well-structured layered modules as failing: stable leaves (constants, error types, I≈0) and unstable orchestrators (`main.rs`, bootstrap/wiring code, I≈1) both failed even when architecturally intentional. This is a known limitation of using raw instability alone — Martin's own metric family pairs it with a second axis, **Abstractness (A)**, and gates on **Distance from the Main Sequence** (`D = |A + I - 1|`) instead.
- Added `mdg.abstractness` (fraction of a module's type declarations that are abstract — trait/interface/protocol vs. concrete struct/class/enum) computed from the UAST, and gate on `mdg.main_sequence_distance` instead of raw `mdg.instability` whenever abstractness is available. A concrete, unstable orchestrator (`main.rs`: I≈1, A≈0) now sits *on* the main sequence (D≈0) and passes — no per-language filename exemption list needed.
- Distance alone doesn't fix the "stable concrete leaf" case (constants.rs: I≈0, A≈0 still gives D=1) — that's Martin's accepted "Zone of Pain" exception. Added a symmetric `is_stable_leaf_module` role exemption (declarations-only module, no branching control flow), mirroring the existing entrypoint-exemption pattern in `file_roles.py`/`gates.py`.
- Scoped to Python, Rust, Go, and TypeScript (their UAST mappers now populate `typeKind`). JavaScript (no abstract-type concept in the language) and C++ (independent, pre-existing mapper bug — real tree-sitter-cpp node names were never wired up) keep the original instability-band gate completely unchanged, so files/languages without abstractness data see byte-identical behavior to before this change.
- `main_sequence_distance_max` (0.5) and `stable_leaf_instability_max` (0.05) are explicitly provisional first-pass thresholds, flagged for a follow-up empirical calibration pass (the existing `[0.3, 0.7]` band came from a PyPI-corpus ECDF calibration this repo doesn't have local access to reproduce).

Closes #124.

## Test plan

- [x] New `tests/graphs/uast/test_abstractness.py` — per-language `typeKind` extraction (Python ABC/Protocol/abstractmethod, Rust trait vs. struct/enum, Go interface vs. struct, TypeScript interface/abstract class/class/enum) and `calculate_abstractness` aggregation.
- [x] New `tests/evaluation/test_file_roles.py` — `is_stable_leaf_module` predicate (declarations-only passes, branching/loop bodies correctly excluded, trivial-return functions still qualify).
- [x] Extended `tests/evaluation/test_calibration.py` with the orchestrator-passes / leaf-fails-without-exemption / leaf-passes-with-exemption / tangled-module-still-fails / distance-boundary / fallback-untouched cases.
- [x] Extended `tests/evaluation/test_gate_parity.py`'s characterization grid with cases straddling `main_sequence_distance_max` and `stable_leaf_instability_max`.
- [x] Full suite: `pytest tests/` — 695 passed, 10 skipped, no regressions.
- [x] `ruff check topos tests` / `ruff format --check topos tests` — clean (matches CI).
- [x] Manual end-to-end check via `CharacteristicMorphism.classify_detailed` with synthetic `main.rs`/`constants.rs`/tangled-module/`main.js` morphisms confirming the exact before/after behavior described above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **COMPOSABLE no longer gates raw instability alone for languages with Abstractness support**: the `mdg.instability` gate (fixed `[0.3, 0.7]` band) flagged well-structured layered modules as failing — stable leaves (constants, error types, I≈0) and unstable orchestrators (`main.rs`/bootstrap wiring, I≈1) both got penalized even when architecturally intentional, because a raw-instability band ignores Robert Martin's own second axis. `Φ_COMPOSABLE` now pairs instability with a new `mdg.abstractness` metric (fraction of a module's type declarations that are abstract — trait/interface/protocol vs. concrete struct/class/enum) and gates on Distance from the Main Sequence (`mdg.main_sequence_distance = |A + I - 1|`, threshold `≤ 0.5`) instead, whenever abstractness is available. A concrete, unstable orchestrator now sits on the main sequence (D≈0) and is not penalized. Added a symmetric role-based exemption (`is_stable_leaf_module`: a declarations-only module with no branching control flow) for the "stable concrete leaf" case, which distance alone doesn't resolve — mirroring Martin's own accepted "Zone of Pain" exception. Scoped to Python, Rust, Go, and TypeScript files with countable type declarations; JavaScript and C++ (whose UAST mapper has an independent, pre-existing type-declaration bug) keep the original instability-band gate unchanged. `main_sequence_distance_max` (0.5) and `stable_leaf_instability_max` (0.05) are first-pass provisional thresholds, not yet run through the PyPI corpus ECDF calibration the other COMPOSABLE constants received. Closes [#124](https://github.com/Krv-Labs/topos/issues/124).
+
 ## [0.3.10] - 2026-07-11
 
 ### Added
--- a/docs/uast-industry-standards.md
+++ b/docs/uast-industry-standards.md
@@ -170,7 +170,19 @@ export interface MethodDeclNode extends UNodeBase {
 
 export interface TypeDeclNode extends UNodeBase {
   kind: "TypeDecl";
-  typeKind: "class" | "struct" | "enum" | "interface" | "typeAlias" | "union";
+  typeKind:
+    | "class"
+    | "struct"
+    | "enum"
+    | "interface"
+    | "typeAlias"
+    | "union"
+    | "trait"
+    | "abstractClass"
+    | "protocol";
+  // "trait" | "abstractClass" | "protocol" are abstract; the rest are
+  // concrete. Drives Martin's Abstractness metric (mdg.abstractness,
+  // topos/functors/probes/uast/abstractness.py) — see issue #124.
   name: IdentifierNode;
   members: UNode[];
   bases?: TypeRefNode[];
--- a/tests/evaluation/test_calibration.py
+++ b/tests/evaluation/test_calibration.py
@@ -172,3 +172,63 @@ def test_composable_high_instability_still_fails_without_entrypoint_flag():
         fan_out=2.0,
     )
     assert decision.achieved is False
+
+
+# ---------------------------------------------------------------------------
+# Distance-from-Main-Sequence gate (issue #124) — abstractness present
+# ---------------------------------------------------------------------------
+
+
+def test_composable_orchestrator_passes_via_distance_without_entrypoint_flag():
+    # I=1, A=0 (concrete, unstable orchestrator, e.g. main.rs) sits on the
+    # main sequence (D=0) — should pass without needing is_entrypoint_module.
+    decision = score_coupling(instability=1.0, abstractness=0.0, fan_in=0.0)
+    assert decision.achieved is True
+    assert "within tolerance" in decision.interpretation["mdg.main_sequence_distance"]
+    # Informational only — mdg.instability itself is not gated here, but is
+    # still surfaced so users see why a "too high" reading isn't a failure.
+    assert "too high" in decision.interpretation["mdg.instability"]
+
+
+def test_composable_stable_leaf_fails_distance_without_exemption():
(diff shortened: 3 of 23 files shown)


jeremy-wayland:

PR #161, opened 2026-07-13
fix: make ast.max_function_complexity language-neutral

### LINKED ISSUE #153: SIMPLE: ast.max_function_complexity is vacuous (always 0) for every language except Python
Found while porting `topos/evaluation/characteristic_morphism.py` to Rust (v0.4.0 migration, issue #144), tracing what `ast.max_function_complexity` actually does.

`topos/functors/probes/ast/complexity.py::calculate_function_complexities` finds function nodes via `ast.nodes_of_type("function_definition", "async_function_definition")` — these are **Python-specific tree-sitter native node kind names**. For each one it builds a per-function sub-`ProgramObject`:

```python
func_ast = ProgramObject(root=node, source=ast.source, language=ast.language)
```

This does not pass `uast_root=...`, so `func_ast.uast_root` stays at the dataclass default of `None`. Inside `_calculate_node_complexity`:

```python
def _calculate_node_complexity(ast: ProgramObject) -> int:
    if ast.uast_root is not None:
        return _calculate_cyclomatic_complexity_uast(ast.uast_root)   # <- language-neutral path, never reached
    decision_count = 0
    for node in ast.traverse():
        if node.type in DECISION_NODE_TYPES:   # <- Python-only native node-type strings
            ...
```

Since `uast_root` is always `None` here, the language-neutral `DECISION_UAST_KINDS` branch (`IfStmt`/`ForStmt`/`WhileStmt`/`MatchStmt`/`TryStmt` — clearly intended to be the multi-language path) is dead code. The function always falls through to `DECISION_NODE_TYPES`, which is a Python-specific list (`if_statement`, `elif_clause`, `except_clause`, `list_comprehension`, ...).

Impact: for every non-Python language, `ast.nodes_of_type("function_definition", "async_function_definition")` finds **zero** functions (Rust's function node kind is `function_item`, not `function_definition`; similarly for JS/TS/Go/C++), so `calculate_function_complexities` returns `{}` and `calculate_max_function_complexity` returns `0` — always within the `<= 10.0` gate, a silent vacuous pass. `ast.max_function_complexity` is effectively Python-only today.

Suggested fix: wire `_calculate_cyclomatic_complexity_uast` up for real — walk the file's already-built UAST for `FunctionDecl`/`MethodDecl` nodes directly (no per-function `ProgramObject` reconstruction needed) instead of re-querying native tree-sitter node types. The Rust port (crates/topos-core/src/functors/probes/ast/complexity.rs, landing alongside this issue) does exactly that as the from-scratch implementation, since porting the Python bug faithfully seemed worse than fixing it — worth backporting the same fix to Python for parity, or superseding Python entirely once the Rust core is wired into topos-mcp (#148).

jeremy-wayland:
## Summary
- `calculate_function_complexities`/`calculate_function_complexity_entries` in `topos/functors/probes/ast/complexity.py` found function nodes via Python-specific tree-sitter native node type strings (`function_definition`/`async_function_definition`), so Rust (`function_item`), Go (`function_declaration`), and JS/TS (`function_declaration`/`arrow_function`/etc.) matched **zero** nodes — `calculate_max_function_complexity` silently returned `0`, always within the `<= 10.0` SIMPLE gate. A vacuous pass, not a real evaluation.
- The per-function code also rebuilt a fresh `ProgramObject(root=node, source=ast.source, language=ast.language)` without `uast_root=...`, so the already-correct, language-neutral `_calculate_cyclomatic_complexity_uast` path (walking `IfStmt`/`ForStmt`/`WhileStmt`/`MatchStmt`/`TryStmt`) was dead code for every language, including Python.
- Fix: both functions now walk the file's already-built UAST directly for `FunctionDecl`/`MethodDecl` nodes whenever `ast.uast_root` is populated — which it always is for any `ProgramObject` built via `ProgramMorphism`/`parse_source` (the real production path for every supported language). No native-node re-querying, no per-function `ProgramObject` reconstruction.
  - Name extraction: `UASTNode` has no stored `name` attribute, so a function's identifier is recovered by scanning its own children for a name-shaped *native* token kind (`identifier`, `property_identifier`, `field_identifier`, `type_identifier`, `package_identifier` — tree-sitter-javascript, for instance, names a method's identifier `property_identifier`, not `identifier`, so this had to be checked per-mapper rather than assumed).
  - `async` detection: tree-sitter models `async` as an unnamed leading token, which is filtered out of the UAST entirely — but it's still part of the declaration node's own byte span, so a text sniff at the start of that span recovers it uniformly (Python `async def`, JS/TS `async function`/`async () =>`/`async foo()`, Rust `async fn`).
  - Scope chain (for `qualified_name`/`method`/`closure` classification) is now a top-down recursive walk carrying an ancestor stack, since `UASTNode` has no parent pointer (unlike tree-sitter's `.parent`).
  - Anonymous callables with no bound identifier of their own (e.g. a JS arrow function assigned via `const f = (x) => {...}`, where the name lives on the enclosing `variable_declarator`) are skipped — matching the pre-existing behavior of never picking up unnamed callables (Python lambdas were never matched either).
- A Python-only native tree-sitter fallback is kept, used only when `ast.uast_root is None` (i.e. a `ProgramObject` constructed directly rather than via `ProgramMorphism`/`parse_source`) — real callers never hit it, but it's guarded by its own regression test rather than left as untested dead code.

## Test plan
- [x] `tests/functors/probes/ast/test_function_complexity_entries.py`: updated to build the `ProgramObject` via `ProgramMorphism` (exercising the real, now-fixed production path) instead of constructing it directly; all prior Python assertions (kind, span, nesting, gate-metric parity) still hold unchanged. Added a dedicated test for the `uast_root=None` legacy fallback path.
- [x] New `tests/functors/probes/ast/test_function_complexity_multilang.py`: reproduces the reported bug for Rust, Go, JavaScript, and TypeScript — a branchy function (`if`/`if`/`for`+nested `if`/`match`-or-`while`) that must score complexity `6`, not the prior vacuous `0`/`1`; also covers method/closure/async_function kind classification across all four languages, plus an end-to-end check through `ASTRepresentation.metrics()["ast.max_function_complexity"]` (the exact gate metric).
- [x] Full `pytest` (688 passed, 10 skipped) and `ruff check` / `ruff format --check` clean.
- [x] No Rust (`src/`/`crates/`) changes needed — this is a Python-side fix; the Rust port referenced in the issue is a separate from-scratch implementation.

Fixes #153

### DIFF
--- a/tests/functors/probes/ast/test_function_complexity_entries.py
+++ b/tests/functors/probes/ast/test_function_complexity_entries.py
@@ -2,10 +2,16 @@
 
 Covers nested functions / closures, methods, and module-level code so the
 ``ast.max_function_complexity`` gate can always be mapped back to a span.
+
+``_entries`` builds the ``ProgramObject`` via ``ProgramMorphism`` (rather
+than constructing it directly), so these tests exercise the real production
+code path -- ``uast_root`` populated -- which is what actually drives
+``ast.max_function_complexity`` for real callers (see issue #153).
 """
 
 from __future__ import annotations
 
+from topos.core.morphism import ProgramMorphism
 from topos.core.object import ProgramObject
 from topos.functors.probes.ast.complexity import (
     calculate_function_complexity_entries,
@@ -15,7 +21,7 @@ from topos.utils.tree_sitter import parse_python
 
 
 def _entries(code: str):
-    obj = ProgramObject(root=parse_python(code), source=code, language="python")
+    obj = ProgramMorphism(source=code, language="python").ast
     return {e.qualified_name: e for e in calculate_function_complexity_entries(obj)}
 
 
@@ -70,7 +76,7 @@ def test_nested_closure_is_dotted_and_outer_includes_nested() -> None:
 
 def test_module_level_only_has_no_function_entries() -> None:
     code = "x = 1\nif x:\n    y = 2\nelse:\n    y = 3\n"
-    obj = ProgramObject(root=parse_python(code), source=code, language="python")
+    obj = ProgramMorphism(source=code, language="python").ast
     assert calculate_function_complexity_entries(obj) == []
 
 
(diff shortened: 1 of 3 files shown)


jeremy-wayland:

PR #162, opened 2026-07-13
fix(secure): bridge DDG statement/sub-expression granularity mismatch in taint_flow_paths

### LINKED ISSUE #154: SECURE: taint_flow_paths can't connect DDG edges to source/sink nodes (granularity mismatch)
Found while porting `topos/functors/probes/cpg/taint.py` to Rust (v0.4.0 migration, issue #145).

`taint_flow_paths` builds DDG adjacency from `topos/graphs/pdg/object.py`'s `DependenceEdge`s, whose `source`/`target` are **whole-statement** UAST node ids (see `_compute_data_dependence`: it operates on `statements: list[UASTNode]`, the top-level nodes CFG blocks record — an `AssignExpr` or `ExprStmt` as a single node, never a sub-expression).

Source/sink detection in the same function matches individual **sub-expression** nodes instead — a bare `Identifier`/`MemberExpr` whose sliced source text equals a registry entry (`"input"`, `"eval"`, ...):

```python
if node.kind in ("Identifier", "MemberExpr") and snippet in source_registry:
    sources.append(nid)
```

These two id spaces essentially never intersect. For the textbook case:

```python
def f():
    x = input()
    eval(x)
```

the DDG edge is `(id_of_"x = input()", id_of_"eval(x)", var="x")` — both statement-level ids. But the taint source detected is the **identifier** `input`'s own id (a different, nested node), which is never a key in the DDG adjacency (`forward`, built from statement-level edges only). So `bfs_reachable(forward, src)` from that identifier's id finds nothing, and `taint_flow_paths` returns 0 for this and presumably most realistic source→sink patterns.

Verified in the Rust port (`crates/topos-core/src/functors/probes/cpg/taint.rs`, ported faithfully) — `direct_source_to_sink_flow_is_not_detected_today` reproduces it as a documented-limitation test rather than a bug fix, since untangling the statement-vs-subexpression granularity mismatch is a real design change to either the PDG or the taint probe's source/sink matching, not a quick fix suitable for a straight port.

Suggested fix direction: either (a) have `_compute_data_dependence` record def/use at the identifier-node granularity instead of statement granularity, or (b) have the taint probe treat a source/sink as "reachable" if it's a descendant of (or equal to) a DDG-connected statement, not just an exact id match.

jeremy-wayland:
## Summary

Fixes #154 — `taint_flow_paths` built DDG adjacency from `topos/graphs/pdg/object.py`'s statement-level `DependenceEdge`s, but detected sources/sinks at sub-expression granularity (a bare `Identifier`/`MemberExpr` for sources, a `CallExpr` for sinks). These two id spaces essentially never intersected, so even the issue's textbook case:

```python
def f():
    x = input()
    eval(x)
```

returned `0` taint flows instead of `1`.

- `taint_flow_paths` (`topos/functors/probes/cpg/taint.py`) now bridges the gap by **containment**: each detected source/sink resolves to the smallest DDG-participating statement whose byte span contains it (falling back to itself when it already *is* a DDG-participating statement, so previously-working flat cases are unchanged). This is a single up-front pass building `{ddg_stmt_id: span}`, then one pass per source/sink to find its nearest enclosing statement — not an O(nodes × ddg_endpoints) scan.
- Per the issue's own guidance, this stays scoped to the taint probe: `_compute_data_dependence`, `DependenceEdge`, and the statement-granularity DDG model in `topos/graphs/pdg/object.py` are untouched. No PDG redesign.

### A necessary prerequisite bug found while reproducing end-to-end

Reproducing the issue's exact snippet initially still returned `0` even with the containment fix, because **no `DATA` dependence edge was ever produced at all** for `x = input(); eval(x)` — a separate, deeper bug from the one described in the issue. `_identifier_name` (`topos/graphs/pdg/object.py`) fell back to each `Identifier` occurrence's own node id whenever no `"name"` attribute was present — which is always, since no UAST mapper populates one. Since a node's id is a hash of `(lang, kind, start_byte, end_byte, parent_id)`, the two distinct occurrences of `x` (the def in `x = input()` and the use in `eval(x)`) never shared a key, so `_compute_data_dependence`'s `last_def` bookkeeping could never link them. This affected every language, not just Python (confirmed with `x = 1\ny = x + 2\n` producing 0 `pdg.data_deps` before this fix).

Since the granularity-bridging fix has nothing to bridge without a real DDG edge to begin with, I threaded an optional `source: str = ""` parameter through `ProgramDependenceGraph.from_uast` (and its two call sites: `topos/graphs/cpg/builder.py::build_cpg` and `topos/core/morphism.py::build_pdg`), so `_identifier_name` can recover real token text via span-slicing (mirroring the existing `CodePropertyGraph.node_text` pattern) instead of relying on the id fallback. When no source is given, behavior is unchanged (verified with a dedicated regression test). This is a minimal, additive change — no changes to `_compute_data_dependence`'s statement-walking algorithm, `DependenceEdge`, or the overall DDG/CDG model.

## Test plan

- [x] New `tests/functors/probes/cpg/test_taint.py`:
  - `test_taint_flow_paths_direct_python_source_to_sink` — issue #154's exact repro, end-to-end (parse → UAST → CFG → PDG → CPG → `taint_flow_paths` >= 1).
  - `test_taint_flow_paths_direct_go_source_to_sink` — same granularity fix for a non-Python language (`os.Getenv` → `exec.Command`).
  - `test_taint_flow_paths_no_flow_when_source_and_sink_unrelated` — a source and sink with no real data dependence between them still count 0.
  - `test_taint_flow_paths_clean_code_has_no_flows`
  - `test_taint_flow_paths_respects_allowlist`
- [x] New tests in `tests/graphs/pdg/test_pdg_program_dependence.py`:
  - `test_pdg_emits_data_dependence_across_statements` — guards the identifier-name prerequisite fix.
  - `test_pdg_data_dependence_without_source_falls_back_gracefully` — confirms the no-`source` path is unchanged (no data-dependence edges, no crash) for direct `ProgramDependenceGraph.from_uast` callers that don't pass source.
- [x] Existing security/CPG/PDG tests unaffected — checked `tests/graphs/cpg/test_cpg.py`, `tests/graphs/pdg/test_pdg_program_dependence.py`, `tests/evaluation/test_suppression.py`, `tests/mcp/test_evaluate.py`, `tests/evaluation/test_gate_parity.py` for any hardcoded exact `cpg.taint_flows`/`pdg.data_deps` counts derived from real code (none found — all such assertions build `ClassificationResult`/scoring inputs manually, or only check dangerous-call counts which this change doesn't touch).
- [x] Full `pytest`: 678 passed, 10 skipped.
- [x] `ruff check` and `ruff format --check`: clean.
- [x] Confirmed no Rust duplicate: `crates/topos-core/src/functors/probes/cpg/taint.rs` mentioned in the issue does not exist on `main` (it's from an unrelated draft migration branch) — nothing to touch there.

### DIFF
--- a/tests/functors/probes/cpg/test_taint.py
+++ b/tests/functors/probes/cpg/test_taint.py
@@ -1,0 +1,67 @@
+"""
+Regression tests for ``taint_flow_paths`` (issue #154).
+
+DDG edges connect whole *statement* nodes (see
+``ProgramDependenceGraph._compute_data_dependence``), while source/sink
+detection matches individual *sub-expression* nodes (a bare
+``Identifier``/``MemberExpr`` for sources, a ``CallExpr`` for sinks). These
+two id spaces essentially never intersect directly, so ``taint_flow_paths``
+used to return 0 even for the textbook direct-flow case. It now bridges the
+gap by containment: a source/sink is "connected" if it is a descendant of
+(or equal to) a DDG-participating statement.
+"""
+
+from __future__ import annotations
+
+from topos.core.morphism import ProgramMorphism
+from topos.functors.probes.cpg.taint import taint_flow_paths
+from topos.graphs.cpg.object import CodePropertyGraph
+
+
+def _cpg(source: str, language: str = "python") -> CodePropertyGraph:
+    morphism = ProgramMorphism(source=source, language=language)
+    cpg = morphism.build_cpg()
+    assert cpg is not None
+    return cpg
+
+
+def test_taint_flow_paths_direct_python_source_to_sink():
+    """Issue #154's exact repro: `x = input(); eval(x)` must count >= 1."""
+    cpg = _cpg("def f():\n    x = input()\n    eval(x)\n")
+    assert taint_flow_paths(cpg) >= 1
+
+
+def test_taint_flow_paths_direct_go_source_to_sink():
+    """Same granularity-bridging behavior for a non-Python language."""
+    source = (
+        "package main\n\n"
+        "import (\n"
+        '\t"os"\n'
+        '\t"os/exec"\n'
+        ")\n\n"
+        "func run() {\n"
+        "\tvar v string\n"
+        '\tv = os.Getenv("X")\n'
+        '\texec.Command("sh", "-c", v)\n'
+        "}\n"
+    )
+    cpg = _cpg(source, language="go")
+    assert taint_flow_paths(cpg) >= 1
+
+
+def test_taint_flow_paths_no_flow_when_source_and_sink_unrelated():
+    """A source and a sink with no data-dependence between them stay at 0."""
+    cpg = _cpg("def f():\n    x = input()\n    y = 1\n    eval(y)\n")
+    assert taint_flow_paths(cpg) == 0
+
+
+def test_taint_flow_paths_clean_code_has_no_flows():
+    cpg = _cpg("def f(a, b):\n    return a + b\n")
+    assert taint_flow_paths(cpg) == 0
+
+
+def test_taint_flow_paths_respects_allowlist():
+    """An allowlisted sink pattern is excluded even when a real flow exists."""
+    cpg = _cpg("def f():\n    x = input()\n    eval(x)\n")
+    assert taint_flow_paths(cpg) >= 1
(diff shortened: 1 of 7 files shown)


jeremy-wayland:

PR #160, opened 2026-07-13
fix(uast): correct C++ mapper declaration/expression node names

jeremy-wayland:
## Summary

- `mapper_cpp.py`'s `_DECLARATION_TYPES` was copy-pasted from the Python/Rust mappers and used node names that don't exist in the real `tree-sitter-cpp` grammar (`class_definition`/`struct_item`/`enum_item` instead of the real `class_specifier`/`struct_specifier`/`enum_specifier`). Every C++ class/struct/enum/union silently fell through to `Unknown` in the UAST — verified empirically against the vendored grammar. Two `_EXPRESSION_TYPES` entries (`assignment`, `subscript`) had the same problem; real names are `assignment_expression`/`subscript_expression`.
- Replaced with the verified real grammar node names (`class_specifier`, `struct_specifier`, `enum_specifier`, `union_specifier` → `TypeDecl`).
- `declaration` (free-standing) and `field_declaration` (class members) are dual-purpose in the grammar — a variable/member declaration and a declaration-only function/method signature (forward declarations, pure-virtual methods) share the same wrapper node. Disambiguated by whether a `function_declarator` child is present.
- Added pure-virtual-method detection (`virtual ... = 0;`) so a class/struct containing one classifies as `typeKind=abstractClass` — this makes Martin's Abstractness metric (#124, PR #157) available for C++ as a fast-follow. This PR only fixes the mapper itself; wiring `"cpp"` into `_ABSTRACTNESS_SUPPORTED_LANGUAGES` is a one-line change deferred until `topos/graphs/uast/object.py` lands on `main` via #157 (it doesn't exist there yet).
- Control-flow/statement node names (`if_statement`, `for_statement`, `while_statement`, etc.) were already correct — this is specifically a declaration/type-counting fix, not a CFG/cyclomatic-complexity fix.

Note: this branch independently duplicates the small `extract_attributes` plumbing addition to `mapper_common.py` that PR #157 also makes (both branch off `main`, which doesn't have it yet). Whichever of #157/#158 merges second will hit a trivial, easily-resolved conflict on that one function signature.

Closes #158.

## Test plan

[...]

PR #139, opened 2026-07-12
Methods Upgrade: unified refactoring suite (cycles, dependencies, process)

jeremy-wayland:
## Summary

Implements the Methods Upgrade milestone (closes #83, #84, #86) as one unified refactoring-guidance suite instead of three disconnected features. Supersedes drafts #88 and #89 (closed).

- **`src/ph.rs`** (new): fundamental cycle-basis extraction (spanning tree + back-edge closure) on the CFG — points cyclomatic complexity's count at the actual loops/branches and maps each cycle generator to its source line range (issue #83).
- **`src/frc.rs`** (new): shared Forman-Ricci curvature engine —
  - `balanced_forman_curvature` (Topping et al., ICLR 2022) applied to the MDG, naming concrete dependency edges worth strengthening (issue #84).
  - `directed_forman_curvature` (Samal et al.) applied to GitNexus process graphs, finding execution "choke points" (issue #86).
- **`topos/graphs/process/`** (new): `ProcessGraph` representation over GitNexus `Process`/`STEP_IN_PROCESS` data.
- **CLI**: `topos refactor cycles|dependencies|process` (three subcommands, each calling straight into the Python probes).
- **MCP**: one `topos_refactor(target="cycles"|"dependencies"|"process", ...)` tool — collapsed from an initial three-tool design specifically to fit under the MCP tool-surface wire-size ratchet (`tests/mcp/test_context_budget.py`) after merging v0.3.10's own schema growth; full design rationale (including why the tool is unified while the CLI stays split) lives in `docs/refactor-suite.md`.

All of this is **purely advisory** — none of it feeds SIMPLE/COMPOSABLE/SECURE scoring, consistent with issue #86's explicit acceptance criterion applied across the whole suite.

Version bumped to `0.3.11` (next in the 0.3.x cadence after this branch's merge of the `v0.3.10` release).

## Test plan

- [x] `cargo test` — 35 passed (includes a hand-derived exact curvature value, bottleneck/choke-point detection on synthetic graphs, and the `betti_1`/cyclomatic-complexity cross-check)
- [x] `cargo clippy -- -D warnings` / `cargo fmt --check` — clean
- [x] `pytest` — 698 passed, 11 skipped (opt-in perf benchmark)
- [x] `ruff check` / `ruff format --check` — clean
- [x] `python scripts/check_versions.py` — passes at 0.3.11
- [x] Confirmed zero diff in `topos/evaluation/`/`topos/core/` (no scoring regression)
[...]

PR #163, opened 2026-07-13
Release v0.3.11 Consolidation + Review Fixes

### LINKED ISSUE #83: [methods] PH Homology with Cycle Detection -> Code Snippets
## Idea

Currently we use **cyclomatic complexity** as a summary statistic for a CFG. Instead, let's implement **persistent homology** with a **betweenness-centrality filtration** (important nodes appear first).

From the persistence computation we can recover the **cycle generators** ($H_1$ representatives) and point agents directly to the **code snippets** corresponding to those loops — enabling targeted suggestions for improving simplicity by eliminating loops.

## Motivation

- Cyclomatic complexity is a single scalar; it tells you *how many* independent loops exist but not *where* they are or *which* matter most.
- A betweenness-centrality filtration orders nodes so structurally important control-flow nodes enter first, surfacing the most significant cycles early in the persistence diagram.
- Cycle generators give concrete, localizable evidence: each $H_1$ class maps back to specific CFG nodes -> source lines.

## Proposed approach

1. Build the CFG (already available via the SIMPLE generator pipeline).
2. Compute **betweenness centrality** per node; use it (or its inverse) as the filtration value.
3. Run **persistent homology** ($H_1$) over the filtration.
4. Extract **cycle generators** for persistent $H_1$ classes.
5. Map generator nodes back to source ranges and emit **code snippets** agents can act on.

## Deliverables

- [ ] Betweenness-centrality filtration over the CFG
- [ ] $H_1$ persistent homology + cycle generator extraction
- [ ] Mapping from cycle generators to source code snippets
- [ ] Integration as a summary statistic alongside / replacing cyclomatic complexity

### LINKED ISSUE #84: [methods] Curvature for dependency bottleneck detection
## Goal

Augment the **COMPOSABLE** pillar with an **edge-level, cycle/redundancy-aware** structural signal: compute **Forman-Ricci Curvature (FRC)** over the Module Dependency Graph (MDG) that `GitNexus` already produces, and use negative-curvature edges to flag **dependency bottlenecks** and point agents at the specific dependency to cut.

> Note: FRC is **local** (1-hop), not global — it is *not* a replacement for global metrics. Its value over Martin instability is that it is **edge-level** and **triangle/redundancy-aware**, not that it is global.

## Why this differs from Martin instability

Martin instability `I = Ce / (Ca + Ce)` is a **node**-level **degree ratio** (in/out coupling counts). Two failure modes it can't distinguish:

- The **plain** Forman curvature reduces to \`4 - deg(u) - deg(v)\` — pure endpoint degree, i.e. largely **redundant** with Martin. We must use the **augmented (triangle) form** to add information Martin lacks.
- With the triangle term, FRC separates a **bridge** edge (tree-like, no shared neighbors -> very negative) from a **redundantly-connected** edge of *identical degree* (many shared neighbors / cycles -> positive). Martin scores both the same.
- FRC is **edge-level**: it localizes to the specific import/dependency to refactor, not just "module X is unstable."

## Implementation

Port the FRC implementation from [aidos-lab/curvature-filtrations](https://github.com/aidos-lab/curvature-filtrations/blob/main/scott/geometry/measures/forman.py), which **uses triangle information** into Rust.

### Notes
- **Weights (optional):** edge weight = coupling strength / call frequency from the GitNexus trace; node weight = internal complexity (LOC or cyclomatic, already computed for SIMPLE). Start unweighted (augmented) for a first cut.
- **Ranking:** most-negative ~5-10% of edges = bottleneck candidates.
- **Optional — balanced Forman** (Topping et al., *Understanding Over-Squashing and Bottlenecks on Graphs via Curvature*, ICLR 2022): the reference file also implements \`balanced_forman_curvature\`, a stronger bottleneck/over-squashing detector. Consider as a follow-up if augmented Forman is too coarse.

## Integration

- **\`topos evaluate\`**: report "Structural Bottlenecks" when an edge's FRC falls below a threshold, with a concrete reason (e.g. *"This import is a negative-curvature bottleneck (no shared dependencies) between Module A and Module B"*).
- **MCP / agents**: expose per-edge curvature so agents can target it directly (*"Find the MDG edge with the most negative Forman-Ricci curvature and propose an interface-based decoupling to raise its curvature toward zero"*).
- **\`topos inspect-graph\`** (optional): return JSON of the MDG + per-edge curvature so agents can "see" topology.

## Scope note

If a genuinely **global** bottleneck signal is wanted later, pair FRC with **edge betweenness** (Girvan-Newman) or a **spectral** measure (Fiedler value / Cheeger). Out of scope for this issue.

## Deliverables

- [ ] Port augmented Forman-Ricci (unweighted + weighted) from aidos-lab into \`src/\` (NetworkX)
- [ ] Build MDG from GitNexus output as a NetworkX graph
- [ ] Edge ranking + threshold for bottleneck flagging
- [ ] COMPOSABLE metadata surfacing bottlenecks with human-readable reasons
- [ ] \`topos evaluate\` integration
- [ ] (optional) \`balanced_forman_curvature\` variant
- [ ] (optional) \`topos inspect-graph\` JSON output for agents

### LINKED ISSUE #86: [methods] Refactor Tool for Process Graphs using Curvature
# Issue #86: [Refactor Tool] Integrate Process Graphs and Discrete Curvature for Architectural Analysis

## Objective
Implement **Process Graph Analysis** as an interactive refactoring tool (following the design of MDG/CFG interactive tools). This tool reads execution flows and `STEP_IN_PROCESS` relationships from `.gitnexus/lbug` to identify structural bottlenecks and provide targeted decoupling suggestions.

## Core Deliverable: Directed Forman-Ricci Curvature (FRC)
We will implement a directed version of **Forman-Ricci Curvature** (FRC) to compute curvature values for every edge $e = (u \to v)$ participating in a process pathway:

$$
\text{Ric}(e) = w_e \left( \frac{w_u}{w_e} + \frac{w_v}{w_e} - \sum_{e_{\text{in}} \sim u} \sqrt{\frac{w_u}{w_{e_{\text{in}}}}} - \sum_{e_{\text{out}} \sim v} \sqrt{\frac{w_v}{w_{e_{\text{out}}}}} \right)
$$

- **Bottleneck Detection:** Highly negative curvature edges ($\text{Ric}(e) \ll 0$) indicate funnels where independent execution paths squeeze through a single transition.
- **Action/Suggestion:** Prompt the user to use **Asynchronous Decoupling** (e.g., event pub-sub or message queues) between $u$ and $v$ to break synchronous coupling.

## Future Suggestions (Extensions)
- **1-Cochain Hodge Decomposition:** Analyze flow cycles/loops for potential Dependency Inversion (DIP).
- **Effective Resistance ($R_{\text{eff}}$):** Measure serial path fragility to suggest Concurrency/Parallelization or Caching.

### 1. Data Ingestion (`topos/graphs/process/`)
- Create/extend `ProcessGraph` or `ModuleDependencyGraph` to parse `Process` nodes and `STEP_IN_PROCESS` relationships from `.gitnexus/lbug`.
- Map 1-indexed `step` properties to order execution paths.

### 2. Directed FRC Engine (`topos/functors/probes/process/`)
- Implement the directed Forman-Ricci curvature algorithm over process graphs.
- Support fast execution ($O(|E|)$ space/time complexity).

### 3. CLI Interactive Interface (`topos/cli/commands/refactor.py`)
- Add `topos refactor process <filepath>` to:
  1. Ingest process graphs intersecting target file.
  2. Compute FRC along the pathways.
  3. List highly negative curvature transitions (bottlenecks).
  4. Display actionable decoupling suggestions.

## Acceptance Criteria
- Process graph evaluations **must not** alter core SIMPLE, COMPOSABLE, or SECURE medal scoring.
- Curvature calculation on $|V| = 10,000$, $|E| = 50,000$ must run in **<100ms**.
- `topos refactor process` lists bottlenecks and refactoring suggestions.
- Unit tests cover directed FRC on synthetic quivers and cycles.

### LINKED ISSUE #124: COMPOSABLE pillar: instability band penalizes intentional layering in Rust crates

### LINKED ISSUE #127: Generalize AST test/source code separation across languages

### LINKED ISSUE #151: SIMPLE/Rust: cfg.cyclomatic doesn't count match-arm branches

### LINKED ISSUE #152: SIMPLE: ast.entropy (gzip-ratio proxy) penalizes small, dense-but-simple functions

### LINKED ISSUE #153: SIMPLE: ast.max_function_complexity is vacuous (always 0) for every language except Python

### LINKED ISSUE #154: SECURE: taint_flow_paths can't connect DDG edges to source/sink nodes (granularity mismatch)

### LINKED ISSUE #158: C++ UAST mapper declaration node names don't match tree-sitter-cpp grammar (TypeDecl/VarDecl unrecognized)
## Summary

`topos/graphs/uast/mapper_cpp.py`'s `_DECLARATION_TYPES` (and a couple of `_EXPRESSION_TYPES` entries) were copy-pasted from the Python/Rust mappers and use node type names that don't exist in the real `tree-sitter-cpp` grammar. As a result, **every C++ class/struct/enum/union declaration currently maps to `Unknown` in the UAST**, independent of any other feature — verified empirically by parsing sample C++ with `tree-sitter-cpp` directly.

Found while scoping issue #124 / PR #157 (COMPOSABLE Abstractness metric) — C++ was excluded from that work specifically because this bug means C++ type-declaration counting doesn't work at all today.

## Current (wrong) mapping

```python
_DECLARATION_TYPES = {
    "function_definition": "FunctionDecl",  # correct (coincidentally)
    "class_definition": "TypeDecl",   # real name: class_specifier
    "struct_item": "TypeDecl",        # real name: struct_specifier
    "enum_item": "TypeDecl",          # real name: enum_specifier
    "impl_item": "TypeDecl",          # not a C++ concept (Rust-ism)
    "function_item": "FunctionDecl",  # not a C++ concept (Rust-ism)
    "method_definition": "MethodDecl",     # not a real tree-sitter-cpp node name
    "lexical_declaration": "VarDecl",      # not a real tree-sitter-cpp node name (JS-ism)
    "variable_declaration": "VarDecl",     # not a real tree-sitter-cpp node name
}

_EXPRESSION_TYPES = {
    ...
    "assignment": "AssignExpr",   # real name: assignment_expression
    "subscript": "MemberExpr",    # real name: subscript_expression
}
```

`union_specifier` (C++ unions) has no entry at all.

## Verified real tree-sitter-cpp grammar node names

Confirmed by parsing sample C++ (classes with virtual/pure-virtual methods, structs with base classes, `enum class`, `union`, free functions, local/global variable declarations, member access, subscripts, assignments) directly with the `tree-sitter-cpp` grammar already vendored in this repo:

| Concept | Real node name |
|---|---|
| class | `class_specifier` |
| struct | `struct_specifier` |
| enum | `enum_specifier` |
| union | `union_specifier` |
| function / method | `function_definition` (top-level functions **and** in-class method definitions both use this node — methods are not distinguished from free functions today) |
| pure-virtual / declaration-only method | `field_declaration` containing a `function_declarator` (no body) — e.g. `virtual double area() const = 0;` |
| variable declaration (local or global) | `declaration` |
| class member field | `field_declaration` (also used for declaration-only methods, see above — needs disambiguation by checking for a `function_declarator` child) |
| assignment | `assignment_expression` (not `assignment`) |
| subscript | `subscript_expression` (not `subscript`) |

Statement-level control flow (`if_statement`, `for_statement`, `while_statement`, `try_statement`, `throw_statement`, `return_statement`, `expression_statement`, `call_expression`, `binary_expression`, `field_expression`, `unary_expression`) already use names that happen to match the current mapper, so CFG-driven metrics (`cfg.cyclomatic`, branch counting) are **not** believed to be affected — this is specifically a declaration/type-counting gap (`TypeDecl`, `VarDecl`, and the two listed `_EXPRESSION_TYPES` entries).

## Impact

- Any UAST-derived declaration count for C++ (`uast_kind_histogram`, `structural_summary`'s `declaration_count`) undercounts — all classes/structs/enums/unions and most variable declarations are invisible, bucketed as `Unknown`.
- Blocks adding Martin's Abstractness metric (`mdg.abstractness`, #124/#157) for C++ — there's no `typeKind`-classifiable `TypeDecl` node to attach `trait`/`interface`/`class`/`struct`-style classification to.
- `assignment_expression`/`subscript_expression` fall to generic/`Unknown` handling instead of `AssignExpr`/`MemberExpr`.

## Proposed fix

1. Replace `_DECLARATION_TYPES` with the real node names above (`class_specifier`, `struct_specifier`, `enum_specifier`, `union_specifier` → `TypeDecl`; `declaration` → `VarDecl`; keep `function_definition` → `FunctionDecl`; drop the Rust-isms `impl_item`/`function_item`).
2. Fix the two `_EXPRESSION_TYPES` entries (`assignment_expression`, `subscript_expression`).
3. Add an `extract_type_attributes` hook (matching the pattern already used in `mapper_rust.py`/`mapper_python.py`/`mapper_go.py`/`mapper_typescript.py` for #124) that detects pure-virtual methods (`field_declaration` with a `function_declarator` and a `= 0` pure-specifier) to classify a `class_specifier`/`struct_specifier` as `typeKind="abstractClass"`, enabling C++ Abstractness support as a fast-follow to #124/#157.
4. Add C++ to `_ABSTRACTNESS_SUPPORTED_LANGUAGES` in `topos/graphs/uast/object.py` once (3) lands.
5. Regression test asserting `TypeDecl`/`VarDecl` nodes are actually produced for representative C++ source (nothing today catches that they're all `Unknown`).

Related: #124, PR #157.

sgathrid:
## Summary

Builds the v0.3.11 release queue by merging the reviewed PR branches onto `origin/main`, then applies the fixes needed to make the combined branch mergeable and release-ready. Closes #83, closes #84, closes #86, closes #124, closes #127, closes #151, closes #152, closes #153, closes #154, closes #158.

> [!IMPORTANT]
> **This pulls in changes from all `v0.3.11` open PRs. The below should be closed when this PR is merged!**

Included PRs:

- #133 — test-node filtering interface
- #157 — COMPOSABLE gate main-sequence distance / abstractness
- #160 — C++ UAST mapper declaration-name fixes
- #155 — Python/Rust match-arm CFG counting
- #156 — entropy size-floor blend
- #161 — language-neutral `ast.max_function_complexity`
- #139 — refactoring suite: cycles, dependency curvature, process curvature
- #162 — DDG/sub-expression taint-flow granularity fix

## Fixes applied on top of the merged queue

| Branch / PR | Issue found during queue review | Fix applied in this release branch |
| --- | --- | --- |
| `review-pr133` / #133 | The Python test-node filtering test still printed a debug S-expression. | Removed the debug import/print so the test stays quiet. |
| `review-pr133` / #133 | A `__main__` guard with an `else`/`elif` was dropped wholesale, silently removing real fallback logic. | Only bare `__main__` guards are treated as scaffolding; guards with alternatives remain in the UAST. |
| `review-pr156` / #156 | The entropy size-floor blend was symmetric: it could hide genuinely low-entropy tiny repetitive code by lifting the score toward neutral. | Made the correction asymmetric: tiny-input blending only lowers suspiciously high zlib-overhead ratios; low-entropy ratios remain raw. |
| `review-pr157` / #157 | `is_stable_leaf_module` treated any no-branching code as a frozen leaf, so straight-line executable side-effect code could receive the COMPOSABLE exemption. | Tightened the predicate to reject calls/functions/methods, limiting the exemption to declarations-only stable leaves. |
| `review-pr157` / #157 | The entrypoint exemption also applied to main-sequence distance, allowing abstract unstable entrypoints in the Zone of Uselessness to pass. | Restricted the entrypoint carve-out to the raw-instability fallback and pinned distance-mode behavior with a regression test. |
| `review-pr160` / #160 | C++ function-pointer declarations such as `void (*handler)(int);` were misclassified as `FunctionDecl`, which would corrupt C++ abstractness once enabled. | Kept function-pointer declarations as `VarDecl`; then enabled C++ abstractness after the mapper was safe. |
| `review-pr155` / #155 | A comment between a Rust `match` scrutinee and its arm block shifted the block out of a fixed child slot and collapsed the match to one edge. | Locate the arm container by node kind rather than fixed position. |
| `review-pr161` / #161 | The new UAST function-complexity path undercounted Python-specific decision forms and still returned vacuous `0` for C++ functions whose names are nested under declarator wrappers. | Preserved Python's native complexity path and added recursive name recovery for C++ declarator wrappers. |
| `review-pr139` / #139 | Legacy/non-numeric GitNexus process `step` properties could raise raw `ValueError`; unexpected `topos_refactor` exceptions escaped the structured MCP result shape. | Fall back to discovery order for non-numeric `step` values and return structured `RefactorResult.error` output for unexpected refactor-tool exceptions. |
| `review-pr139` / #139 | CLI hotspots passed unsupported `evidence` fields that `RefactorHotspot` silently discarded. | Removed the dead fields so CLI construction matches the canonical schema. |
| `review-pr162` / #162 | DDG construction used one file-global reaching-definitions map, so same-named locals could create cross-function taint flows. | Build data-dependence per CFG callable, keeping taint propagation intra-procedural. |

## Release metadata

- Release version is `0.3.11` in `Cargo.toml`, `Cargo.lock`, the VS Code extension package and lockfile, and both version fields in `.mcp/server.json`.
- `pyproject.toml` continues to derive the Python package version from `Cargo.toml` through Maturin.
- `CHANGELOG.md` now marks the queue as the `0.3.11` release dated 2026-07-13.
- `scripts/check_versions.py` now validates the MCP Registry manifest as well as the Python/Cargo and VS Code surfaces, preventing future release-version drift.

## Validation

- `28 passed` across the focused calibration, Python AST filtering, and Rust CFG regression suites.
- `4 passed` in `tests/test_version.py`.
- Ruff passed for all touched Python files.
- `python3 -S scripts/check_versions.py` reports `version check passed (0.3.11)`.
- Cargo metadata resolves `topos-functors` as `0.3.11`.
- `git diff --check` passes.

# MCP Registry release after merge

After this PR is merged and `0.3.11` is available on PyPI:

```bash
mcp-publisher login github
mcp-publisher validate .mcp/server.json
```

> [!NOTE]
> it appears someone changed our GitHub org access policies -- do not want to revert changes that were made for a reason. But if you are having trouble publishing using `mcp-publisher` via GitHub OAuth, try using your gh-cli auth token instead as follows:

```bash
mcp-publisher logout
MCP_GITHUB_TOKEN="$(gh auth token)" mcp-publisher login github
```


jeremy-wayland:
## Dogfooding follow-up

Downloaded the actual v0.3.11 release binary (via `gh run download` on this PR's build) and ran it against every `topos` CLI command with real fixtures across Python/Rust/C++. Found and fixed two bugs plus one doc drift, all merged into `release-0.3.11`:

- **CHANGELOG drift**: the `#124` entry said C++ still gets the raw instability-band gate, but this same release also merges the `#158`/`#160` C++ mapper fix, which enables C++ abstractness too. Corrected.
- **COMPOSABLE scored 0% for isolated files**: `calculate_coupling`'s "no signal" fallback (`instability=0.5` for a file with zero measured coupling) sat in the optimal band under the old raw-instability gate, but combined with the common `abstractness=0.0` case it landed `main_sequence_distance` exactly on its ceiling — passing the hard gate at the boundary while scoring `0%`, showing `FAIL` in the CLI despite counting toward an `IDEAL` badge. Distance mode now requires a real measured coupling signal.
  → [#164](https://github.com/Krv-Labs/topos/pull/164)
- **GitNexus branch-scoped `.gitnexus` stores silently served the wrong branch's data**: GitNexus writes the first `gitnexus analyze` to a flat `.gitnexus/lbug` slot, but re-running `analyze` on a different branch creates `.gitnexus/branches/<branch>-<hash>/lbug` instead of updating the flat slot — and Topos had zero awareness of that layout. Worse, `generate_depgraph()` always overwrote Topos's own freshness fingerprint at the flat slot regardless of where the store actually landed, so after switching branches it could report "fresh" while silently loading the *other* branch's stale graph. Added branch-aware store resolution (`resolve_lbug_store`, `current_git_branch`) used everywhere `.gitnexus` gets read, plus a new `branch_not_indexed` state when the current branch genuinely isn't indexed.
  → [#165](https://github.com/Krv-Labs/topos/pull/165)

Both merged; a follow-up `ruff format` pass (pure formatting, no logic changes) landed in `a4792e5`.

jeremy-wayland (approved):
lfg

krv-labs-ai-developer (approved):
Verified v0.3.11 binary from release-0.3.11 artifacts. CLI commands successfully tested against both Python (topos/cli/main.py) and Rust (src/lib.rs) source files. No regressions observed.

krv-labs-ai-developer (approved):
Verified PR #166 fixes. The size-aware entropy scoring logic correctly addresses the SIMPLE score-shaping regression for small functions. Verified locally with the latest v0.3.11 binary against src/lib.rs; metrics are now stable and accurate.

PR #166, opened 2026-07-14
fix: entropy size-floor score shaping still penalizes tiny simple code (#152)

jeremy-wayland:
## Summary

Follow-up to PR #156 (entropy size-floor blend), surfaced while smoke-testing the v0.3.11 release binary against issue #152's own repro.

PR #156 fixed the **hard entropy gate** (`ratio <= 0.8`) for tiny dense-but-simple functions — its own pinned Rust test (`test_kolmogorov_proxy_tiny_dense_function_stays_in_band`) only asserts that. It did not fix the **SIMPLE score**, which is what issue #152 actually complained about.

Live-tested `topos evaluate --language rust` against the issue's exact repro:

```
match version   (cyclomatic 8, entropy 0.670): SIMPLE 66% PASS
array version    (cyclomatic 1, entropy 0.738): SIMPLE 52% FAIL
```

The strictly simpler, branch-free refactor still scores *worse* and fails SIMPLE — the exact bug #152 was filed about, just past the hard gate boundary instead of past it.

## Root cause

`score_simple`'s combined score is `min()` of each metric's quality. `_quality()`'s curve for `ast.entropy` is symmetric around `entropy_ideal` (0.5): `1 - 2*|value - ideal|`. PR #156's blend already pulls the *raw ratio* partway back toward neutral for tiny files, but any partial, continuous blend preserves the original ordering between two below-floor files — it can narrow the gap but never close it. The array version's true raw ratio is inherently higher than the match's (fewer repeated tokens for zlib to exploit), so no amount of *raw-metric* blending alone fixes the score-level regression; the *quality curve* itself needs to know the file is tiny.

## Fix

- `topos/evaluation/policies/calibration.py`: new `SIMPLE.entropy_size_floor_bytes = 200.0`, mirroring Rust's `ENTROPY_SIZE_FLOOR_BYTES`.
- `topos/evaluation/policies/simple.py`: `score_simple`/`_quality` take an optional `source_size_bytes`. Below the floor, an **above-ideal** entropy reading no longer lowers the score (it's measurement noise, not a real "too dense" signal at this size) — but a **below-ideal** reading (repetition/boilerplate) stays fully penalized, since that signal holds at any size. Above the floor, or when the caller doesn't pass a size, behavior is byte-identical to before.
- `topos/evaluation/characteristic_morphism.py`: threads `len(morphism.source.encode("utf-8"))` through to the SIMPLE dispatcher (the other two dispatchers accept-and-ignore it, matching the existing `is_stable_leaf_module` pattern).

Re-tested the same repro end-to-end after the fix:

```
match:  SIMPLE 80%  PASS
array:  SIMPLE 95%  PASS
```

## Incidental fix

`tests/mcp/test_assess.py::test_assess_improvement_has_no_regression_diff` depended on the *same* underlying symmetric-curve quirk in the opposite direction — its sibling helper `_force_regression_score`'s docstring explicitly called out "added branching can score as an improvement under SIMPLE" as a known scoring quirk it had to work around. That quirk is now gone, so the test's original fixture direction (branchy scoring as an "improvement" over a one-liner) no longer holds — correctly. Swapped the fixture direction to a quirk-free simplification (branchy → simple) so the test no longer depends on that behavior either way, and updated the now-stale docstring.

## Validation

- New regression tests in `tests/evaluation/test_calibration.py` pinning: above-floor behavior unchanged, below-ideal still penalized below the floor, above-ideal not penalized below the floor, and the issue's exact match-vs-array pair (array now scores `>=` match).
- Full suite: `804 passed, 11 skipped` (up from 803/11 baseline — 1 net new file's worth of tests, 1 existing test's fixture corrected).
- `ruff format` / `ruff check` clean on all touched files.
- Manual end-to-end re-test via `CharacteristicMorphism` and via a live-built `topos-macos-arm64` binary against both snippets (see numbers above).

### DIFF
--- a/tests/evaluation/test_calibration.py
+++ b/tests/evaluation/test_calibration.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import pytest
 from topos.evaluation.policies.calibration import (
     CLONE,
     COMPOSABLE,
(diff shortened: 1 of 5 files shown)


jeremy-wayland:
Merged into `release-0.3.11` as `982afe4`. Quick summary of the final fix:

PR #156 fixed the hard entropy gate but not the SIMPLE score itself — the `_quality()` curve for `ast.entropy` is symmetric around the ideal (0.5), so a tiny above-ideal reading still dragged the score down even after being blended by size. This PR makes the curve size-aware: below `entropy_size_floor_bytes` (200), an *above*-ideal entropy reading no longer lowers the SIMPLE score (it's measurement noise from zlib's per-stream overhead at that size), while a *below*-ideal reading (real repetition/boilerplate) is still fully penalized.

Verified against issue #152's exact repro on the built `release-0.3.11` binary: match now scores SIMPLE 80% (was 66%), array-lookup now scores 95% (was 52%, FAIL) — the simpler refactor correctly scores better.

PR #172, opened 2026-07-17
fix(vscode): avoid files for failed downloads

### LINKED ISSUE #173: VS Code downloader leaves an empty file after non-200 responses
## Problem
`downloadFile` created the destination stream before confirming an HTTP 200 response. A redirect or non-200 response could therefore leave an empty destination file behind while the function rejected.

## Expected behavior
Failed downloads must not leave a destination file.

## Fix
PR #172 defers stream creation until after the response is confirmed as HTTP 200.

sgathrid:
> [!IMPORTANT]
> **TL;DR:** fix git action failure

## Summary
- defer creation of the download destination stream until the response is confirmed as HTTP 200
- prevent redirects and non-200 responses from creating an empty destination file
- retain cleanup for failures that occur after a successful response starts writing

## Verification
- pnpm run test:unit (24 passed)
- pnpm run check-types
- pnpm run lint
- git diff --check

The pnpm checks ran in a temporary verification copy. The repository continues to use its existing package-lock setup; no pnpm lockfile or workspace configuration is included in this PR.

Closes #173

### DIFF
--- a/extensions/vscode/src/runtime.ts
+++ b/extensions/vscode/src/runtime.ts
@@ -160,11 +160,13 @@ export function downloadFile(
     redirectsRemaining: number = MAX_REDIRECTS
 ): Promise<void> {
     return new Promise((resolve, reject) => {
-        const file = fs.createWriteStream(destPath);
+        let file: fs.WriteStream | undefined;
 
         const cleanup = () => {
-            file.close();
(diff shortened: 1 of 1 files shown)


jeremy-wayland (approved):
👍

PR #174, opened 2026-07-17
fix(security): keep allowlisted `SECURE` surfaces consistent

### LINKED ISSUE #168: cpg.metrics() ignores allowlist while security_findings honors it
## Problem

The two SECURE surfaces disagree when an allowlist is active:

- `security_findings(cpg, allow=…)` filters allowlisted callees via `effective_registry` (see `topos/mcp/security_findings.py` → `_map_sighthound_finding` / `dangerous_call_findings`).
- `CodePropertyGraph.metrics()` → `topos.utils.sighthound.count_findings` has **no** `allow` parameter, so `cpg.dangerous_calls` / `cpg.taint_flows` still count findings whose callee is allowlisted.

Result: for the same code, the SECURE **metric** can be > 0 while the **findings list** excludes that call. This is invisible today and confusing when someone relies on an allowlist to suppress a known-safe pattern.

Note: this is pre-existing (not introduced by #167) — `metrics()` has never accepted an `allow` argument. Surfacing it here because #167 touched the surrounding classification code.

## Scope / decision needed

- Decide whether `metrics()` should accept and thread an allowlist through `count_findings` (changes the public `metrics()` signature and its callers), or whether the metric is intentionally allowlist-agnostic and the divergence should just be documented.
- Reconcile the two SECURE surfaces so their counts can't contradict.

Refs #167.

sgathrid:
## Summary

- preserve `CodePropertyGraph.metrics()` as the canonical raw SECURE representation contract
- compute allowlist-adjusted metrics through the same active engine as raw classification (Sighthound or local probes)
- share suffix-aware allowlist matching across Sighthound counts and diagnostics
- resolve taint findings to the actionable sink operation, including cross-file flows
- add regressions for Sighthound-only findings and unrelated active risks

## Root cause

Raw SECURE metrics used Sighthound when available, but `apply_allowlist()` always recomputed adjusted counts with the smaller local dangerous-API registry. The Sighthound diagnostics path also filtered through that local registry, so a non-allowlisted Sighthound-only finding could remain active while `secure_adjusted` incorrectly passed—or disappear from `security_findings` entirely.

The fix keeps raw scoring canonical and introduces an explicit allow-aware CPG metrics path used only by the disclosed acknowledgement overlay.

## Verification

- `uv run pytest -q` — 817 passed, 11 skipped
- `uv run ruff check topos tests`
- `uv run ruff format --check topos tests`
- `git diff --check`

Fixes #168

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Sighthound/Corgea SECURE Outsourcing:** Deprecates Python-based pattern matching and BFS taint-path tracing on the Code Property Graph (CPG) by outsourcing security analysis directly to Corgea/Sighthound when available on system `PATH`. When `sighthound` is present, Topos invokes it to parse JSON-formatted findings and maps them into standard `cpg.dangerous_calls` and `cpg.taint_flows`. If `sighthound` is absent, Topos gracefully falls back to local CPG danger and taint probes.
 
+### Fixed
+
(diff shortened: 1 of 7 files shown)


jeremy-wayland (approved):
nice fix 🐛

PR #185, opened 2026-07-21
feat: add OpenClaw and Hermes skill definitions for Topos

krv-labs-ai-developer:
Adds OpenClaw skill definition and Hermes skill integration for Topos structural analysis.

PR #187, opened 2026-07-21
fix: address PR #185 review findings

sgathrid:
Companion PR into #185. Fixes the findings from review; merge this, then #185.

## Changes

- **AGENTS.md** — moved the Agent Skills section outside the `OPENWIKI:START/END` managed block into its own `SKILLS:START/END` block (mirroring `OPENWIKI-POLICY`). OpenWiki regenerates the managed region and commits AGENTS.md in its docs PR, so the section would have been deleted on the next run. Content is unchanged.
- **clawhub-publish.yml** — corrected the header comment: without `CLAWHUB_TOKEN`, pushes to `main`/`v*` tags **fail** (upstream exits 1), they do not skip. Noted that fork PRs also fail because fork-originated `pull_request` runs never get `id-token: write`.
- **check_skill.py** — nested metadata checks are now scoped to their actual blocks (a `tags:` under `metadata.openclaw` no longer satisfies the `metadata.hermes.tags` requirement). Removed the `REQUIRED_FRONTMATTER_KEYS` loop (duplicated the `parse_scalar` checks, producing double errors) and the single-use `frontmatter_contains` wrapper.
- **README.md** — dropped a stray blank line.

## Verified

- `python scripts/check_skill.py` passes (1 skill, version 0.3.12).
- Negative test: misplaced keys under the wrong metadata block are now rejected.
- Workflow YAML parses.
- Not changed: `owner: Krv-Labs` casing — the PR's own dry-run (run 29837947288) accepted it and reports the skill `alreadySynced` on clawhub.ai.

### DIFF
--- a/.github/workflows/clawhub-publish.yml
+++ b/.github/workflows/clawhub-publish.yml
@@ -6,7 +6,10 @@
 #   2. GitHub → Settings → Secrets and variables → Actions → New repository secret
 #      Name: CLAWHUB_TOKEN
 #      Value: <paste token>
-# Without this secret, PR dry-runs still work; pushes to main and v* tags skip real publish.
+# Without this secret, PR dry-runs still work, but pushes to main and v* tags FAIL:
+# the upstream workflow exits 1 ("Real skill publishes need secrets.clawhub_token").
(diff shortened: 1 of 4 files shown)


jeremy-wayland (approved):
cheers mate

PR #203, opened 2026-07-23
fix: detect Homebrew installs across platforms

### LINKED ISSUE #200: Tighten Homebrew Cellar path detection in CLI install method
## Summary

`_is_homebrew_executable` treats **any** path segment named `Cellar` as Homebrew. That can mis-detect install method and print the wrong `update` / `uninstall` guidance.

**Scope:** CLI only (`topos/cli/installation.py`). Not related to the Homebrew tap publish workflow.

## What

```mermaid
flowchart TD
  A["detect_install_info()"] --> B{"Python package metadata found?"}
  B -->|yes| C["package-manager / source / etc."]
  B -->|no| D{"_is_homebrew_executable(active_exe)?"}
  D -->|yes| E["method = homebrew<br/>brew upgrade / uninstall"]
  D -->|no| F["binary-installer or unknown"]
```

Current check:

```python
# topos/cli/installation.py
if "Cellar" in path.parts:
    return True
```

| Path | Today | Should be |
| --- | --- | --- |
| `/opt/homebrew/Cellar/topos/0.3.12/bin/topos` | homebrew | homebrew |
| `/usr/local/Cellar/topos/0.3.12/bin/topos` | homebrew | homebrew |
| `$HOMEBREW_PREFIX/bin/topos` | homebrew (via env) | homebrew |
| `/Users/me/projects/Cellar/bin/topos` | **homebrew** | **not** homebrew |
| `/tmp/fake/Cellar/topos/x/bin/topos` | **homebrew** | **not** homebrew |

## Why

| Impact | Detail |
| --- | --- |
| Wrong UX | `topos update` / uninstall may suggest `brew …` for a non-brew binary |
| Severity | Low — needs a path segment literally named `Cellar` |
| Frequency | Rare, but the heuristic is overly broad |

## Suggested fix

Prefer a **prefix + layout** check over a bare name match:

1. Resolve known Homebrew roots: `HOMEBREW_PREFIX`, then common defaults (`/opt/homebrew`, `/usr/local`, Linuxbrew paths if you already support them).
2. Treat as Homebrew only if the resolved exe is under one of those roots **and** either:
   - path matches `…/Cellar/<formula>/<version>/…`, or
   - path is under `<prefix>/bin` (linked keg) / already covered by `is_relative_to(prefix)`.
3. Drop standalone `"Cellar" in path.parts` (or keep it only as a last resort behind a real prefix check).

Pseudo:

```text
homebrew if:
  path under known_prefix AND (
    parts contain Cellar at the right depth, OR
    path.is_relative_to(prefix)
  )
```

## Tests

Existing coverage in `tests/cli/test_helpers.py`:

- `test_detect_install_method_homebrew_cellar`
- `test_detect_install_method_homebrew_prefix_env`
- `test_detect_install_method_homebrew_wins_over_stale_binary_provenance`

Add negatives:

| Case | Expected `method` |
| --- | --- |
| `…/projects/Cellar/bin/topos` (no brew prefix) | not `homebrew` |
| Real Cellar under `/opt/homebrew` | `homebrew` |
| Under `HOMEBREW_PREFIX` only | `homebrew` (unchanged) |

## Out of scope

- Release / tap publish workflow
- Changing brew formula URLs or version normalization
- Broader install-method redesign

## Acceptance

- [ ] No bare `"Cellar" in path.parts` true-positive without a Homebrew root
- [ ] Real Cellar + `HOMEBREW_PREFIX` cases still detect `homebrew`
- [ ] Unit tests for false-positive path above

ColumbusLabs:
## Summary
- detect standard Apple Silicon, Intel macOS, and Linuxbrew Cellar layouts even when `HOMEBREW_PREFIX` is absent
- preserve custom Homebrew-prefix support while ignoring malformed or relative prefix values
- avoid misclassifying direct `/usr/local/bin/topos` binary-installer installs as Homebrew
- add regression coverage for standard, custom, unrelated, malformed, and provenance cases

## Verification
- `pytest -q` (839 passed, 11 skipped)
- `ruff check topos tests`
- `ruff format --check topos tests`
- `cargo test` (39 passed)
- `cargo clippy -- -D warnings`
- `cargo fmt --check`
- independent review completed with no remaining findings

Closes #200

### DIFF
--- a/tests/cli/test_helpers.py
+++ b/tests/cli/test_helpers.py
@@ -230,6 +230,22 @@ def test_detect_install_method_homebrew_cellar(monkeypatch):
         assert info.uninstall_cmd == "brew uninstall topos"
 
 
+def test_detect_install_method_homebrew_apple_silicon_linked_keg(monkeypatch):
+    def raise_not_found(name: str):
+        raise PackageNotFoundError
+
+    with monkeypatch.context() as m:
+        m.setattr("importlib.metadata.distribution", raise_not_found)
+        m.setattr("topos.cli.installation.load_provenance", lambda: None)
+        m.setattr(
+            "topos.cli.installation.active_executable",
+            lambda: Path("/opt/homebrew/bin/topos"),
+        )
+        m.delenv("HOMEBREW_PREFIX", raising=False)
+
+        assert detect_install_info().method == "homebrew"
+
+
 def test_detect_install_method_homebrew_intel_cellar(monkeypatch):
     def raise_not_found(name: str):
         raise PackageNotFoundError
--- a/tests/cli/test_helpers.py
+++ b/tests/cli/test_helpers.py
@@ -230,6 +230,38 @@ def test_detect_install_method_homebrew_cellar(monkeypatch):
         assert info.uninstall_cmd == "brew uninstall topos"
 
 
+def test_detect_install_method_homebrew_intel_cellar(monkeypatch):
+    def raise_not_found(name: str):
+        raise PackageNotFoundError
+
+    with monkeypatch.context() as m:
+        m.setattr("importlib.metadata.distribution", raise_not_found)
+        m.setattr("topos.cli.installation.load_provenance", lambda: None)
+        m.setattr(
+            "topos.cli.installation.active_executable",
+            lambda: Path("/usr/local/Cellar/topos/0.3.12/bin/topos"),
+        )
+        m.delenv("HOMEBREW_PREFIX", raising=False)
+
+        assert detect_install_info().method == "homebrew"
+
+
+def test_detect_install_method_homebrew_linux_cellar(monkeypatch):
+    def raise_not_found(name: str):
+        raise PackageNotFoundError
+
+    with monkeypatch.context() as m:
+        m.setattr("importlib.metadata.distribution", raise_not_found)
+        m.setattr("topos.cli.installation.load_provenance", lambda: None)
+        m.setattr(
+            "topos.cli.installation.active_executable",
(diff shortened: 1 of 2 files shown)


jeremy-wayland:
Thanks @ColumbusLabs for picking up #200 and for the thorough test matrix — the `/usr/local/bin` guard, platform coverage, and malformed-HOMEBREW_PREFIX handling were exactly what we needed as the base.

  I pushed two follow-up commits on top of your branch with a reconciled three-tier _is_homebrew_executable:

  1. HOMEBREW_PREFIX — broad prefix match (your existing behavior)
  2. Private defaults (/opt/homebrew, Linuxbrew) — broad prefix match for linked kegs
  3. `/usr/local` — Cellar layout only, so we don’t misclassify curl/binary installs in /usr/local/bin

  That middle tier came from the parallel work in #204 (@armorbreak001): the insight that paths under Homebrew-only roots like `/opt/homebrew/bin/topos` should count even when HOMEBREW_PREFIX isn’t set. We didn’t merge #204 as-is because applying the same broad rule to `/usr/local reintroduces` a false positive there — but your linked-keg case is now covered by
`test_detect_install_method_homebrew_apple_silicon_linked_keg`.

 @armorbreak001 — if you’re good with it, I’d like to add a Co-authored-by: trailer on the install/test commits that incorporate that idea so you get proper credit on the contributors graph. 
 
 Give this a thumbs up and i'll add this when merging, following a quick review by @sgathrid.

sgathrid (approved):
Approving, cheers 🍻

> [!NOTE]
> Edge case found here --> changes do not resolve the bug _specifically_ for non-homebrew installs on **Intel** MacOS. As such, just opened issue #220 

The Homebrew/Intel MacOS issue was pre-existing on main, not a regression, so putting this PR through!

PR #222, opened 2026-07-24
fix: avoid Intel Homebrew false positives

### LINKED ISSUE #220: Homebrew Detection Misfire on Intel macOS
## Problem

On Intel macOS, Homebrew's default prefix is `/usr/local`, and the standard `brew shellenv` setup exports `HOMEBREW_PREFIX=/usr/local` into the environment. When that variable is set, Homebrew detection in `topos/cli/installation.py` treats *anything* under the configured prefix as a Homebrew install — with no requirement that the file actually live in a Homebrew Cellar layout. Since the curl binary installer also places its executable at `/usr/local/bin/topos`, a binary-installer install on an Intel Mac that also has Homebrew gets misclassified as a Homebrew install.

This is pre-existing behavior on `main` (the old detection had the same broad prefix match), not a regression from #203. #203 fixed the misclassification when `HOMEBREW_PREFIX` is unset; this issue covers the remaining case where it is set.

## How to replicate

1. On an Intel Mac (or simulated in a test), install topos via the binary installer so the executable is `/usr/local/bin/topos` and provenance records `install_method=binary-installer`.
2. Have Homebrew installed with the documented shell setup, so `HOMEBREW_PREFIX=/usr/local` is exported.
3. Ensure no `topos-mcp` Python distribution is importable (so detection falls through to executable/provenance checks).
4. Run any command that calls `detect_install_info()` (e.g. `topos uninstall` or `topos update`).

Detection returns `homebrew` instead of `binary-installer`. Reproduced end-to-end on the #203 branch.

## What this does

- `topos uninstall` and `topos update` point the user at `brew uninstall topos` / `brew upgrade topos`, which fail because the formula was never installed.
- The correct binary-installer provenance (including its recorded install path and PATH-hint cleanup) is ignored, so the user gets no working removal or update path.

## Suggested fix

When the configured `HOMEBREW_PREFIX` resolves to the shared `/usr/local` root, do not use the broad prefix match. Instead, fall through to the same Cellar-layout check that already guards the default `/usr/local` tier. Genuine Homebrew installs on Intel are still detected, because Homebrew links `/usr/local/bin/topos` as a symlink into the Cellar, and the executable path is resolved through symlinks before matching. Custom prefixes and `/opt/homebrew` keep the broad match unchanged.

Add a regression test covering: `HOMEBREW_PREFIX=/usr/local`, executable at `/usr/local/bin/topos`, binary-installer provenance present — expected result `binary-installer`.

## Priority context

Intel macOS moves to Homebrew Tier 3 in September 2026 and becomes unsupported in September 2027, so the affected population is real but shrinking. Low severity, straightforward fix.

Related: #200, #203

pollychen-lab:
## Summary
- avoid treating every executable under `/usr/local` as Homebrew when `HOMEBREW_PREFIX=/usr/local`
- keep genuine Intel Homebrew Cellar installs detected as Homebrew
- add regression coverage for binary-installer provenance under `/usr/local/bin/topos` with the Intel Homebrew prefix exported

Fixes #220.

## Testing
- `uv run pytest tests/cli/test_helpers.py -q`
- `uv run pytest tests/cli -q`
- `uv run ruff check topos/cli/installation.py tests/cli/test_helpers.py`
- `git diff --check`

### DIFF
--- a/tests/cli/test_helpers.py
+++ b/tests/cli/test_helpers.py
@@ -334,6 +334,49 @@ def test_detect_install_method_binary_in_usr_local_bin(monkeypatch):
         assert info.provenance == provenance
 
 
+def test_detect_install_method_binary_in_usr_local_bin_with_homebrew_prefix(
+    monkeypatch,
+):
+    def raise_not_found(name: str):
+        raise PackageNotFoundError
+
+    provenance = {
+        "install_method": "binary-installer",
+        "install_path": "/usr/local/bin/topos",
+    }
+
+    with monkeypatch.context() as m:
+        m.setattr("importlib.metadata.distribution", raise_not_found)
+        m.setattr("topos.cli.installation.load_provenance", lambda: provenance)
+        m.setattr(
+            "topos.cli.installation.active_executable",
+            lambda: Path("/usr/local/bin/topos"),
+        )
+        m.setenv("HOMEBREW_PREFIX", "/usr/local")
+
+        info = detect_install_info()
+        assert info.method == "binary-installer"
+        assert info.provenance == provenance
+
+
+def test_detect_install_method_homebrew_intel_cellar_with_homebrew_prefix(
+    monkeypatch,
+):
+    def raise_not_found(name: str):
+        raise PackageNotFoundError
+
+    with monkeypatch.context() as m:
+        m.setattr("importlib.metadata.distribution", raise_not_found)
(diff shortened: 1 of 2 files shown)


sgathrid (approved):
LGTM

pollychen-lab:
@sgathrid thanks

Issue #216, opened 2026-07-24
[v0.4.0 parity] topos inspect never scores COMPOSABLE (hard-coded mdg=None)


jeremy-wayland:
_From the review of #159 (Rust v0.4.0 migration). Non-blocking._

**Where:** `topos/cli/src/commands/inspect.rs` — `run`

**What:** `run` calls `classify_with_representations(&classifier, &mut morphism, None)` with a hard-coded `None` MDG, so the COMPOSABLE dimension is never attached and never appears in `inspect` output or `--json`. Python's `inspect` took a `--gitnexus-dir` option and ran `run_classify_file` with it, so COMPOSABLE was measured/shown when a `.gitnexus` store existed. The module doc lists only security-findings/suggestions/suppression as out of scope — not COMPOSABLE.

**Impact:** `topos inspect foo.py` in a repo with an indexed `.gitnexus` shows a `composable:` line under Python but never under the Rust CLI, and there's no flag to enable it.

**Fix:** wire an optional `--gitnexus-dir` into `inspect` (reusing `resolve_composable_mdg` like `evaluate` does), or document that `inspect` deliberately drops COMPOSABLE.

jeremy-wayland:
Fixed in PR #208 (commit 1b6928b): `inspect` now resolves the COMPOSABLE MDG via `resolve_composable_mdg` (mirroring `evaluate`), with `--gitnexus-dir` / `--no-composable` flags. Verified end-to-end — `composable:` appears in both text and `--json`. Will auto-close when #208 reaches `main` (it targets `worktree-rust-migration-v0.4.0`, not the default branch).

jeremy-wayland:
Closed via #159 (merged to `main` at 20ff7d0). Squash merge did not auto-close because closing keywords were not in the merge commit body.

**#216**: `topos inspect` / `topos_inspect_code` now resolve COMPOSABLE (#208/#224).

Issue #193, opened 2026-07-22
SIMPLE gate: cfg.cyclomatic aggregate sum vs. per-function threshold mismatch


jeremy-wayland:
## What's the issue?

`cfg.cyclomatic`'s SIMPLE gate uses a flat threshold (15) against a whole-file merged-CFG sum, but the threshold was calibrated for single-function complexity. Any file with several genuinely-simple functions can fail SIMPLE by function count alone. Two other scoring bugs found in the same dogfooding pass (both already fixed) are included for the record.

## Method / function

`evaluation/policies/{gates,calibration,simple}.rs` (SIMPLE gate), `graphs/cfg/builder.rs` (CFG construction)

## Details

Found while dogfooding v0.4.0 against `topos-engine` itself (#104, PR #191).

### 1. `cfg.cyclomatic` aggregate vs. per-function threshold mismatch (open)

`cfg/builder.rs` deliberately keeps `P = 1` across every callable in a file (one synthetic entry/exit), so `cfg.cyclomatic` is a whole-file sum of every function's complexity — mathematically correct as an aggregate (equals summing each function's own McCabe number), but gated against the same flat threshold (15) used for a single function's complexity. `ast.max_function_complexity` already does the right thing here (true per-function max, computed independently in `functors/probes/ast/complexity.rs`) and correctly passes on the same files.

Confirmed by experiment, not just theory: splitting `graphs/mdg/object.rs` (cyclomatic 48) into two files by responsibility dropped it to 25 and 24 — both still over 15. Closing that kind of gap requires fragmenting into unreasonably many small files, not real cleanup.

Net effect: 33/99 files in `topos-engine` fail the SIMPLE gate on `cfg.cyclomatic` alone despite passing `ast.max_function_complexity` and `ast.entropy`.

`Φ_SIMPLE` currently ANDs all three metrics independently (`achieved = (cyclomatic ≤ gate) ∧ (entropy in band) ∧ (max_func ≤ gate)`, tested in `simple.rs`'s `independent_thresholds_each_fail_alone`). Options to reconcile, in increasing order of change:
- Scale the cyclomatic threshold by function count in the file.
- Drop `cfg.cyclomatic` as a hard gate, keep it as a scored/informational metric only, gate on `ast.max_function_complexity` instead.

Deliberately not fixed in PR #191 — changes SIMPLE's meaning for every future evaluation, wanted a separate decision/review rather than folding it into a cleanup pass.

[...]

PR #159, opened 2026-07-13
migration(rust): v0.4.0

jeremy-wayland:
## TL;DR

- **What:** The entire Python Topos implementation is deleted. All computation lives in a Rust workspace (`topos-engine`, `topos` CLI, `topos-mcp`).
- **Proof:** Parity validated against **`topos-mcp==0.3.12`** (latest published Python) — 12/12 corpus files matched (9 exact, 3 allowlisted correctness improvements); ~13–37× per-invocation speedup over Python.
- **Scope:** This PR ships v0.4.0 behavior. Crates.io publish ([#149](https://github.com/Krv-Labs/topos/issues/149)) and remaining Graphify measures ([#150](https://github.com/Krv-Labs/topos/issues/150)) are explicit follow-ups.

**Do not review this as 462 sequential files.** Use the module map and review passes below.

## Reviewer instructions (dogfood + parity)

Detailed step-by-step for reviewers:

1. **Install both CLIs** — reference `topos-mcp==0.3.12` ([PyPI](https://pypi.org/project/topos-mcp/0.3.12/) / [Release v0.3.12](https://github.com/Krv-Labs/topos/releases/tag/v0.3.12)) vs candidate built from this branch (`cargo build --release -p topos`).
2. **Dogfood one real repo per language** with a local agent — [pandas](https://github.com/pandas-dev/pandas) (Python), [Sighthound](https://github.com/Corgea/Sighthound) (Rust), [express](https://github.com/expressjs/express) (JS), [axios](https://github.com/axios/axios) (TS), [cobra](https://github.com/spf13/cobra) (Go), [fmt](https://github.com/fmtlib/fmt) (C++).
3. **Restore parity harness** (`git checkout 5a98781^ -- parity/ scripts/parity_check.py scripts/benchmark_perf.py`), pin reference to `0.3.12`, run parity + wall-clock benchmarks.

See the [**Reviewer instructions: dogfood + parity** comment](https://github.com/Krv-Labs/topos/pull/159#issuecomment-5070598175) for commands, agree/differ table, and report-back checklist.

## How to review

| Pass | Where to look | What to verify |
| --- | --- | --- |
| **1. Engine parity** | [`topos/engine/src/`](topos/engine/src/) (`core/`, `graphs/`, `evaluation/`, `functors/`) | Lattice scoring unchanged except documented fixes ([#151](https://github.com/Krv-Labs/topos/issues/151)–[#154](https://github.com/Krv-Labs/topos/issues/154), match/switch arm counting) |
[...]

PR #224, opened 2026-07-24
fix(v0.4.0): stack safety + MCP interface correctness

sgathrid:
Stacked on `worktree-rust-migration-v0.4.0`. Per-change detail: `CHANGELOG.md`.

**Pass 1 (original):** deeply nested input must never overflow the stack — provably without changing anything else.
**Pass 2 (added after review):** the MCP surface was telling agents things that weren't true, and charging a lot of context to say them.

## Pass 2 — broken → fixed

Every row reproduced live over stdio, before and after.

| Was broken | Fix | Why it mattered |
| --- | --- | --- |
| Docs + prompt taught `{"params": {…}}`; `rmcp` takes flat args | flattened every example; tests pin docs to the router | **no documented call worked** — `unknown field 'params'` |
| Entries behind `metric_locations` skipped functions it couldn't name — on Rust, most `pub fn`s | entries now cover them | a failing gate had **no target to edit**: `file_roles.rs` gate 30, top entry 6 |
| `cfg.cyclomatic` ranked first, but it can't fail a pillar (#193) | gating tier leads, on all 3 surfaces (`refactor_targets`, `next_actions`, `suggestions`) | agents refactored the **wrong metric**: file-wide 80-vs-15 buried a real 14-vs-10 |
| Docs called `cfg.cyclomatic` a gate | relabeled advisory; `achieved` vs `score` spelled out | `discovery.rs` IDEAL at score 0.0 **read as a bug** |
| `verbose: false` didn't drop `raw_metrics` on project rows or assessments | gated there too | **53,766 → 19,857 B** per page (−63%) |
| GitNexus child output passed through raw | capped at 200 B on a char boundary | **kilobytes of Node stack trace** per warning, even on success |
| `topos_inspect_code` never loaded a dep graph | loads it like `evaluate_file` | same file, **two different verdicts** (SIMPLE_SECURE vs IDEAL) |
| `topos inspect --json` printed the graph-gen banner into stdout | child output captured when `--json` | **invalid JSON on first run** in any repo |
| Nothing said which binary a host was running | `topos://build` + stale-server warning + `topos-mcp --version` | dogfooding burned turns on `stat`/`pgrep`/`strings` |

Also: tool count corrected 17 → **18**.

## Breaking for callers

| Change | Restore old behavior |
| --- | --- |
| Responses default to compact — `raw_metrics` behind `verbose`, null/empty omitted | `verbose: true`; omission is unconditional |
| `refactor_targets` defaults to `3`, was `0` | pass `0` |
| `refactor_targets[0]` is the top gating target when no `preferences.ranking` given | supply a ranking |
| `topos_inspect_code` scores COMPOSABLE (**closes #216**) | `no_composable: true` |
| `worst_files` is `{filepath, lattice_element}`, not full row clones | read the matching `files` row |
| New `binding_constraint`; `serverInfo.version` gains `+build.<epoch>` | additive |

## Pass 1 — stack safety (unchanged from first review)

UAST clone/drop/equality, CFG construction, callable discovery, and PDG dataflow are now iterative; the CFG builder is an explicit task machine; CPG node collection went quadratic → linear. Edge shapes for all six languages are locked as golden contracts captured from the pre-rewrite builder, so drift fails loudly.

Closes #226, #227, #228, #229. #230 (Go's extra module-level callable) stays open — the contracts intentionally lock pre-rewrite behavior.

> [!NOTE]
> Base is `worktree-rust-migration-v0.4.0`, so issues auto-close via commit keywords when the stack lands on `main` through #159 — not from this body.

## Verify

`cargo test --workspace` — **399 passed**, 1 ignored (342 at first review). Clippy `-D warnings` and `fmt --check` clean. Payload measured against a baseline rebuilt from the pre-change tree in a throwaway worktree, not asserted.

## Deferred

| Issue | Why not here |
| --- | --- |
| #231 `cfg.nesting_depth` doesn't reset at branch joins (1008 for a 460-line file) | diagnostic-only; needs post-dominator traversal on the CFG this stack just froze |
| #232 `scores` and `pillars` can disagree on project rows | needs a canonical-value decision |
| #233 four duplicated `agent_contract` builders | refactor, no behavior gain |
| #234 project contract routes to `inspect_code`, not the ROI head | needs a definition of ROI that survives review |
| #235 UAST has no visibility concept | multi-mapper engine feature |

Rejected from dogfooding: anti-fragmentation scoring guardrails (re-litigates #193 backwards, fights the `extract_helper` suggestions) and signature-freeze gates ("without explicit instruction" is unimplementable; the buildable half needs #235).


jeremy-wayland (approved):

jeremy-wayland:
Great updates! Thanks for the thorough review @sgathrid 🙏

PR #246, opened 2026-07-25
Fix: `v0.4.1` installer failures

### LINKED ISSUE #247: `curl` install in `README.md` not working
> [!IMPORTANT]
> **Being addressed by PR #246** -- still pending review

### What happened?

Install command via curl does not work.

### Steps to reproduce

1. run `curl -fsSL https://docs.krv.ai/topos/install.sh | sh`
2. see issue + no install:
```
sh: line 278: syntax error near unexpected token `<'
```

### Expected behavior

proper install

### Version

v0.4.1

sgathrid:
> [!IMPORTANT]
> **No more broken installs!** Fixes three independent install failures — MCP `@mcp` resolution, `curl | sh` parse error, and macOS binaries linking Homebrew OpenSSL. Closes #247.

## 1. MCP server install via `@mcp` (VS Code)

`server.json` used `registryBaseUrl: "https://pypi.org"`, matching the official MCP registry example. The VS Code `@mcp` installer passes it through as `uvx --index-url <registryBaseUrl>`, hitting the PyPI HTML frontend instead of the Simple API — a false "not found in the package registry" even though the package was live.

- `registryBaseUrl` → `https://pypi.org/simple`
- Documented the `/simple` requirement for future publishers
- Fixed relative SVG badge links in `README.md` so they render off-GitHub

**Verified:** `uvx --index-url https://pypi.org/simple topos-mcp@0.4.1` resolves; VS Code `@mcp topos` installs clean.

## 2. `curl | sh` install — closes #247

`preflight_existing_installs` fed a read loop from `done < <(...)`. Process substitution is a bashism POSIX mode rejects, and `/bin/sh` on macOS is bash 3.2 in POSIX mode, so the script failed to parse before running a line:

```
sh: line 278: syntax error near unexpected token `<'
```

- Feed the loop from a here-doc — parses everywhere, and keeps the loop body in the current shell where a pipe would lose `foreign`/`same`/`foreign_paths` to a subshell
- `|| true` on the command substitution, so `set -e` cannot turn a non-zero discovery exit into a *silent early exit* on machines with no existing install
- Advertised command → `| bash` in all six places. The script has a bash shebang and uses arrays and `[[ =~ ]]`, so `| sh` only ever worked on macOS by accident; where `/bin/sh` is dash it never parsed. Fixes Debian/Ubuntu too.

**Verified:** `sh -n`, `bash -n`, bash 5.3 `-n`, and a real `cat install.sh | sh` all pass. The three new tests fail against `d8c007f` and pass here.

## 3. macOS binaries linked Homebrew OpenSSL

`topos/engine/build.rs` emitted `cargo:rustc-link-lib=ssl`/`crypto` with no `static=` prefix (rustc defaults `-l` to dynamic) and probed `brew --prefix openssl@3` for the search path, baking an absolute Homebrew path into every macOS artifact. The hardened runtime implies library validation, so dyld refused to map a dylib with a foreign Team ID and every `topos` call aborted.

**The premise was false.** lbug 0.17.1 statically links mbedtls — `liblbug.a` has zero OpenSSL symbols and zero OpenSSL strings, and neither released binary imported anything from libssl/libcrypto. Dead weight, but dyld loads `LC_LOAD_DYLIB` eagerly, so it still killed startup. build.rs is deleted, which also removes a build-time `brew` shell-out that made artifacts depend on the builder's Homebrew state.

Linux escaped only because the Debian/Ubuntu toolchain passes `--as-needed` by default (GNU ld itself defaults to `--no-as-needed`). A distro default, not a guarantee — so Linux is checked too.

**Blast radius** — every channel served the same artifact: curl installer, Homebrew tap (formula shas come from `checksums.txt`), and the VS Code VSIX (bundles a byte-identical copy, sha `b16be83f…`). The PyPI macOS wheel broke differently: ad-hoc signed, so library validation never applied and it failed only for users *without* openssl@3.

### Guards added

| Guard | Catches |
| --- | --- |
| macOS + Linux linkage checks, before signing | non-portable artifact never gets signed or notarized |
| Smoke test staged binary | outright missing dylib |
| **Smoke test signed binary** | **this bug** — see below |
| macOS wheel linkage check | wheel regressions (Linux covered by auditwheel) |

The signed smoke test is the one that matters. Library validation applies *only* to the signed artifact, and notarization does not detect broken dylib dependencies. Proven by re-signing the broken v0.4.1 binary ad-hoc: it loads the same Homebrew dylib and runs fine. That is also **why this shipped** — `cargo build` produces ad-hoc-signed binaries that work perfectly, so it was unreproducible locally.

**Verified:** all four artifacts rebuild with only `/usr/lib` dependencies and zero OpenSSL imports; the macOS wheel rebuilt via maturin exactly as CI does; the fixed binary runs when signed with Developer ID + hardened runtime where v0.4.1 aborts (134); every guard tested against both the real broken artifact and the fixed one to rule out false positives. 400 Rust + 16 packaging tests pass.

Notarization was validated end-to-end against this fix on a local build — submission `dbb6d47d` returned `status: Accepted`. That was a dry run for confidence only; the artifact users receive is rebuilt, re-signed, and re-notarized by CI on the release tag.

### Rejected alternatives

- **`com.apple.security.cs.disable-library-validation`** — would load the dylib but weakens security *and* leaves users without Homebrew openssl broken ("no such file"). Treats the symptom.
- **`-dead_strip_dylibs`** — masks the root cause; ld64's man page warns against it for dylibs needed via initializers.
- **Static/vendored OpenSSL** — unnecessary, nothing uses a single OpenSSL symbol.

### Tagged release needed
- Merging rebuilds nothing; macOS users stay broken until a tag ships.
- Binaries installed before this carry an embedded upgrade hint reading `| sh`; self-resolves after one upgrade.
### Publish to MCP Registry
```bash
mcp-publisher login github 
##or, if you are getting a permissions issue re organization access to krv labs:
MCP_GITHUB_TOKEN="$(gh auth token)" mcp-publisher login github
```

```bash
```
### Review
- `brew install openssl@3` remains in the wheel job with no known consumer; left because the maturin Linux path isn't verifiable locally.


jeremy-wayland (approved):
LGTM! Thanks for catching this — my dogfooding was all local cargo build, which turns out to ad-hoc-sign and dodge the exact library-validation issue that broke the real curl/Homebrew/VSIX artifacts. The joy of many install channels 🙄 The lbug pin and OpenSSL handling looked ugly but I was too keen once I got my local setup working. This is much cleaner, cheers @sgathrid

PR #251, opened 2026-07-26
Restore CLI UI/UX!

sgathrid:
## Summary

- restore quiet, cumulative `topos evaluate` output with TTY progress on stderr
- add bounded `--info` weak-file investigation, actionable `inspect` output, and a lightweight project config selector
- align help and supporting CLI commands with the same minimal terminal language
- refresh the Rust CLI documentation and remove stale Python-era references

> [!NOTE]
> Seems the previous CLI/Terminal UI changes were lost in the rust refactor. This PR re-introduces proper CLI usage so CLI-native users opting out of MCP can have a first-class experience. _Also just looks noice._

### Before

Directory evaluation streamed every file, verdict, pillar result, and raw metric to stdout:

<details>
<summary>Example legacy output</summary>

```text
./benchmarks/.cache/pypi/sources/botocore/1.43.39/botocore-1.43.39/tests/unit/test_errorfactory.py
  Verdict: 🥈 SIMPLE_SECURE
    simple: 🥉 SIMPLE [46%]
    secure: 🥉 SECURE [100%]
  Raw metrics:
    ast.entropy: 0.231
    ast.max_function_complexity: 2.000
    cfg.cyclomatic: 12.000
    cfg.essential: 1.000
    cfg.longest_path: 2.000
    cfg.nesting_depth: 0.000
    cpg.dangerous_calls: 0.000
    cpg.taint_flows: 0.000
    mdg.abstractness: 0.000
    pdg.control_deps: 0.000
    pdg.data_deps: 23.000
    pdg.density: 0.719

./benchmarks/.cache/pypi/sources/botocore/1.43.39/botocore-1.43.39/tests/unit/test_eventstream.py
  Verdict: 🥈 SIMPLE_SECURE
    simple: 🥉 SIMPLE [22%]
    secure: 🥉 SECURE [100%]
  Raw metrics:
    ast.entropy: 0.236
    ast.max_function_complexity: 7.000
    cfg.cyclomatic: 31.000
    cfg.essential: 2.000
    cfg.longest_path: 6.000
    cfg.nesting_depth: 78.000
    cpg.dangerous_calls: 0.000
    cpg.taint_flows: 0.000
    mdg.abstractness: 0.000
    pdg.control_deps: 5.000
    pdg.data_deps: 64.000
    pdg.density: 0.651

./benchmarks/.cache/pypi/sources/botocore/1.43.39/botocore-1.43.39/tests/unit/test_exceptions.py
  Verdict: 🥉 SIMPLE
[...]

PR #254, opened 2026-07-27
refactor(cli): split evaluate/inspect into modules

jeremy-wayland:
## Summary
- Split the restored `evaluate` / `inspect` CLI surface into directory modules (`evaluate/{mod,info,info_render,summary}`, `inspect/{mod,detail}`)
- Keep shared terminal chrome in `commands/render.rs` (`paint`, truncates, guides, spinner)
- Update `composable`’s `include_str!` path for the moved inspect entrypoint

Stacked on #251 (`codex/restore-cli-terminal-ux`).

## Test plan
- [x] `cargo check -p topos`
- [x] `cargo test -p topos` (42 passed)
- [x] `cargo fmt -p topos --check`
- [ ] Spot-check `topos evaluate` / `topos inspect` text output still matches #251 UX

Made with [Cursor](https://cursor.com)

PR #255, opened 2026-07-27
cli: combine --priority and --preferences into one flag

jeremy-wayland:
## Summary

Follow-up to #251's review discussion: `--priority` and `--preferences` were doing the same job — `priority` is provably just `ranking[0]` with the rest inferred — so this collapses them into a single `--priority` flag that accepts either form:

- `--priority secure` — a single pillar, same as before
- `--priority secure,simple,composable` — a full ranking, same as before

Applies to `topos evaluate --priority` and `topos config set --priority`, and to the `.topos.toml` `[evaluation] priority` key (now a string or array instead of two separate keys).

## Bonus fixes (found during #251's review)

Splitting `priority`/`preferences` into two flags had produced two divergent "single pillar" code paths, both of which discarded information they shouldn't have:

- **`topos config` interactive TTY selector** used to reset the full preference ranking to the default order whenever you just picked a priority, silently discarding any ranking set earlier via `config set --preferences`. It now reorders the existing ranking around the new priority, same as the scriptable path always did.
- **`topos evaluate --priority`** used to drop the project's persisted preference ranking entirely for that run's refactor-target ordering (`--info`/`--failures --info`), falling back to the default order instead of reordering around the given priority. It now reorders it via the existing `focused_ranking` helper.

Both are covered by new unit tests (`resolved_ranking_preserves_existing_order_around_the_new_priority`, `single_priority_flag_reorders_persisted_preferences_instead_of_dropping_them`, etc.).

## Other changes

- `docs/source/cli.rst`, `index.rst`, `preferences.rst`, `skills/topos/SKILL.md` updated for the single flag.
- `extensions/vscode`: the "Evaluate Project" command now passes `--priority` instead of `--preferences` (setting id `topos.evaluatePreferences` kept as-is to avoid breaking existing user settings.json).

## Verification

- `cargo fmt --all --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace` (444 passed, 1 ignored)


topos/engine/src/config.rs
```diff
@@ -157,12 +157,11 @@ pub fn load_topos_config(start: &Path) -> ToposConfig {
...
-        .and_then(|table| non_empty_str(table.get("priority")))
-        .and_then(|value| parse_priority(&value));
-    let preferences = evaluation
-        .and_then(|table| table.get("preferences"))
-        .and_then(parse_preferences);
+    let priority_value = evaluation.and_then(|table| table.get("priority"));
+    let priority = priority_value
+        .and_then(toml::Value::as_str)
+        .and_then(parse_priority);
+    let preferences = priority_value.and_then(parse_preferences);
```
sgathrid: **Bug (high):** rankings are only taken from `priority` when it is an array:

```rust
let preferences = priority_value.and_then(parse_preferences);
```

That matches the PR/docs *target* schema, but `write_settings` in `topos/cli/src/commands/config.rs` still writes:

- `priority = "<pillar>"` (string)
- `preferences = ["…", "…", "…"]` (legacy key)

So after `config set --priority a,b,c`, load returns `preferences: None`. Existing dual-key / preferences-only files also break silently.

**Load order to implement:**

1. If `priority` is an array → full ranking.
2. Else if `priority` is a string → single-pillar priority.
3. If ranking still empty → legacy `evaluation.preferences` (one-release compat).

Add a test for legacy `preferences` and/or a CLI load-after-write round-trip.

topos/cli/src/commands/config.rs
```diff
@@ -381,4 +386,37 @@ mod tests {
         assert!(updated.contains("preferences = [\"secure\", \"simple\", \"composable\"]"));
```
sgathrid: **Bug (high):** this assertion documents the broken dual-key write shape that the new engine loader no longer reads:

```rust
assert!(updated.contains("priority = \"secure\""));
assert!(updated.contains("preferences = [\"secure\", \"simple\", \"composable\"]"));
```

`write_settings` (unchanged in this PR, ~L242–247) always writes string `priority` + `preferences` array. Engine tests in this PR expect `priority = ["…"]` instead.

**Canonical write (match docs + loader):**

- Always persist the full ranking the CLI already computes as  
  `priority = ["secure", "simple", "composable"]`
- Do **not** write `preferences`; remove the key if present so one source of truth remains.
- Single-pillar `config set --priority secure` still has a full ranking in memory via `resolved_ranking` — persist that array (first element is effective priority).

Replace this test with a **load-after-write** check via `load_topos_config`, and drop the dual-key string asserts.

sgathrid (changes requested):
## Review

Thanks for collapsing `--priority` / `--preferences` — the evaluate flag path and the interactive `resolved_ranking` fix look right.

### Blocking: config writer/loader schema mismatch

This PR changes the **loader** so rankings only come from `evaluation.priority` when it is an array, and docs describe a single `priority` key (string | array). But **`write_settings` is unchanged** and still persists the dual-key shape from #251:

```toml
priority = "secure"
preferences = ["secure", "simple", "composable"]
```

Consequences:

1. **`topos config set --priority composable,secure,simple` does not round-trip.** Ranking is written under `preferences`, then dropped on load (`preferences: None`; `config show` → `preferences  not set`).
2. **Existing `.topos.toml` files** with legacy `preferences = [...]` silently lose rankings (no fallback).
3. **Tests encode the split rather than catch it** — engine tests assert the new array form; the CLI write test still asserts dual keys; there is no load-after-write test.

Inline notes on the loader change and the CLI test that locks in the old write shape. Please fix before merge (I am applying the fix on this branch).

sgathrid:
Addressed the requested changes on this branch:

1. **Merged #251 (incl. #254 module split)** into this head — import conflict in `evaluate/mod.rs` resolved.
2. **Config schema fix**:
   - `write_settings` now writes canonical `priority = ["…", "…", "…"]` and **removes** legacy `preferences`.
   - `load_topos_config` still reads array/string `priority`, with **legacy `preferences` fallback** when no ranking came from `priority`.
   - Round-trip, migration, and legacy-key tests added; `docs/source/cli.rst` documents the on-disk shape.

Verification: `cargo fmt --check`, `cargo clippy -p topos -p topos-engine --all-targets -- -D warnings`, config + full `topos` bin tests green.

Ready for re-review / merge into #251.

sgathrid (approved):
Re-reviewed after the merge + schema fix.

- Import conflict with #254 resolved cleanly in `evaluate/mod.rs`.
- `write_settings` / `load_topos_config` now agree on single-key `priority` (array ranking), with legacy `preferences` read-only fallback and migration on write.
- Round-trip / legacy tests cover the previous silent-drop failure mode.

LGTM for merge into #251.

topos/engine/src/config.rs
sgathrid: resolved by [86cfde1](https://github.com/Krv-Labs/topos/pull/255/commits/86cfde12d4897f7a25062f860d0cd4b5580a17e8)

topos/cli/src/commands/config.rs
sgathrid: resolved via [86cfde1](https://github.com/Krv-Labs/topos/pull/255/commits/86cfde12d4897f7a25062f860d0cd4b5580a17e8) and [6e557e3](https://github.com/Krv-Labs/topos/pull/255/commits/6e557e3c57a420a5ac98b7093a97e61607e92d63)

Issue #215, opened 2026-07-24
Hardening: resolve_within_root lexical fallback weakens symlink containment


jeremy-wayland:
_From the review of #159 (Rust v0.4.0 migration). Non-blocking (hardening)._

**Where:** `topos/mcp/src/security.rs` — `resolve_within_root`

**What:** `joined.canonicalize().unwrap_or_else(|_| normalize(&joined))`. `canonicalize` only succeeds when the *entire* path exists; for a non-existent leaf it falls back to lexical `normalize`, which does **not** resolve symlinks. Python's `is_within_root` used `path.resolve(strict=False)`, resolving symlinks on the existing prefix even when the final component is missing.

Concrete divergence: root `/proj`, symlink `/proj/link -> /etc`, request for `/proj/link/newfile` (absent) — Python denies (resolves to `/etc/newfile`, outside root); Rust **accepts** (lexically `/proj/link/newfile` starts_with `/proj`).

**Impact:** not exploitable today — all callers only read, so a non-existent target just returns NotFound. But the public containment gate returns `Ok` for a path that escapes the root once a symlinked prefix exists — unsafe for any future write/create caller.

**Fix:** canonicalize the longest existing ancestor and re-append the missing tail before the `starts_with` check (or reject when the existing prefix escapes root).

PR #269, opened 2026-08-03
fix(mcp): harden resolve_within_root against symlink escapes

jeremy-wayland:
## Summary
- Canonicalize the longest existing ancestor, then re-append the missing tail before the root `starts_with` check.
- Closes the gap where `/proj/link/newfile` with `link → /etc` was accepted under lexical normalize.
- Stacked on #268 (`release/044-214-graphify-parser`).
- Closes #215.

## Test plan
- [x] `cargo test -p topos-mcp --lib security`
- [x] Unix symlink fixture denies escape via missing leaf
- [x] Missing in-root leaf still allowed

Made with [Cursor](https://cursor.com)

### DIFF
--- a/topos/mcp/src/security.rs
+++ b/topos/mcp/src/security.rs
@@ -68,18 +68,59 @@ fn normalize(path: &Path) -> PathBuf {
     out
 }
 
-/// Resolve a path (absolute or root-relative) and check it's inside the
-/// root, without reading it. Symlinks are resolved when the path exists.
-pub fn resolve_within_root(filepath: &str) -> Result<PathBuf, String> {
-    let root = resolve_file_root()?;
+/// Resolve symlinks on the longest existing prefix, then re-append any
+/// missing tail — matching Python `Path.resolve(strict=False)`.
+///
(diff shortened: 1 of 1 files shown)


jeremy-wayland:
**Security bug found in review, fixed in 4f33181.**

`resolve_existing_prefix` walked the path backwards from the leaf, popping components until one existed. `Path::file_name()` returns `None` once the last component is `..`, so any `..` anywhere in the request bailed the walk out to whole-path lexical normalize — the exact unsafe fallback this function exists to remove. Confirmed exploitable: `link/subdir/../newfile` through an existing `link` symlink with a missing `subdir` was accepted as inside root while actually resolving outside it (plus a path-duplication bug from the old missing-tail re-append).

Fix: walk components forwards instead, canonicalizing each segment against the already-resolved real prefix; once a segment is missing, nothing after it (including `..`) can be a symlink, so the rest is safely applied lexically. Added a regression test (`symlink_escape_via_dotdot_and_missing_intermediate_is_denied`) covering the exact bypass. Also dropped the now-dead `normalize()` helper it replaced.

Issue #258, opened 2026-07-28
Treat --gitnexus-dir / gitnexus_dir as COMPOSABLE project root


jeremy-wayland:
## Summary

`--gitnexus-dir` / MCP `gitnexus_dir` currently only select the `.gitnexus` **store path**. COMPOSABLE freshness fingerprinting and `gitnexus analyze` (via `topos depgraph generate` / auto-refresh) always use the **CLI process cwd** or **MCP file root** as the project root.

That makes this natural workflow fail or hang:

```bash
# from $HOME — wants to evaluate phil only
topos evaluate ~/krv/open-source/phil/phil/magic -r --language python \
  --gitnexus-dir ~/krv/open-source/phil/.gitnexus
```

Today, Topos still fingerprints/`analyze`s **cwd** (e.g. `$HOME`), so a nested `--gitnexus-dir` does not scope the graph to `phil`. Users/agents must `cd` into the repo (or set `TOPOS_MCP_FILE_ROOT`) instead.

Documented as current behavior in #251 docs; this issue is the intended behavior change.

## Proposed behavior

When `--gitnexus-dir` / `gitnexus_dir` is set:

1. Derive the COMPOSABLE **project root** from that path (typically the parent of `.gitnexus`, or the canonicalized store’s containing repo root).
2. Run freshness fingerprinting and `gitnexus analyze` against **that** root — not process cwd / MCP file root.
3. Keep the existing containment rule: the store must still resolve inside a trusted boundary (for MCP: file root; for CLI: decide whether to allow absolute overrides outside cwd — recommend yes for this use case, with clear docs).
4. Evaluating paths outside that root should remain allowed for SIMPLE/SECURE file scoring, or be clearly rejected with a one-line error.

Without the flag, keep today’s default: project root = CLI cwd / MCP file root, store = `<root>/.gitnexus`.

## Acceptance criteria

- [ ] From `$HOME`, `topos evaluate <repo>/… -r --gitnexus-dir <repo>/.gitnexus` freshness-walks and (re)generates only `<repo>`, not `$HOME`.
- [ ] MCP `gitnexus_dir` uses the same derived project-root rule (still contained under `TOPOS_MCP_FILE_ROOT` unless we explicitly widen that).
- [ ] Spinner / stderr copy distinguish “checking freshness” vs “running gitnexus analyze”.
- [ ] Agent contract + CLI `--help` + skill updated for the new semantics.
[...]

PR #270, opened 2026-08-03
feat: treat --gitnexus-dir / gitnexus_dir as COMPOSABLE project root

jeremy-wayland:
## Summary
- When `--gitnexus-dir` / `gitnexus_dir` is set, COMPOSABLE project root is the store's parent (not CLI cwd / MCP file root).
- CLI allows absolute overrides outside cwd; MCP still requires the derived root under `TOPOS_MCP_FILE_ROOT`.
- Spinner phases distinguish freshness vs `gitnexus analyze`.
- Docs: agent contract + skill updated.
- Stacked on #269 (`release/044-215-symlink-containment`).
- Closes #258.

## Test plan
- [x] `cargo test -p topos-mcp --lib evaluation::` (includes root-derivation tests)
- [x] `cargo test -p topos composable`
- [ ] Dogfood: from `$HOME`, `topos evaluate <repo>/… -r --gitnexus-dir <repo>/.gitnexus` fingerprints only `<repo>`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/docs/source/agents.rst
+++ b/docs/source/agents.rst
@@ -404,9 +404,11 @@ Dependency Graph
    (falling back to file mtimes for graphs built before that marker existed), so
    a regenerate reliably clears ``stale``. Never shells out.
 
-``topos_generate_depgraph({"directory": ..., "force": ...})``
-   Runs ``gitnexus analyze`` and writes ``.gitnexus/``. Side-effecting and
-   approval-gated. Requires the ``gitnexus`` CLI (``pnpm add -g gitnexus  # or: npm install -g gitnexus``).
+``topos_generate_depgraph({"directory": ..., "gitnexus_dir": ..., "force": ...})``
+   Runs ``gitnexus analyze`` and writes ``.gitnexus/``. When ``directory`` is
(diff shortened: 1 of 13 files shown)


jeremy-wayland:
**Security bug found in review, fixed in 204fdc3.**

`resolve_composable_project_root` fell back to plain lexical `..`/`.` normalize whenever the joined `gitnexus_dir` path didn't fully exist (common — the store often isn't generated yet). That never follows symlinks on the part of the path that does exist, so `--gitnexus-dir linkdir/.gitnexus` with `linkdir` a symlink out of the trusted root derived a project root that lexically looked contained under `file_root` but really resolved elsewhere — reintroducing, in a second unshared resolver, the exact class of escape #269 just hardened `resolve_within_root` against. Confirmed exploitable via a reproduction test before the fix.

Fix: same forward-walk approach as #269 — canonicalize each existing segment as we go, fall back to lexical join/pop only once a segment is missing. Kept as a local copy in `evaluation/mod.rs` rather than sharing the helper, so this fix doesn't require rebasing onto #269. Added a regression test (`resolve_mcp_composable_project_root_falls_back_when_override_escapes_via_symlink`).

PR #276, opened 2026-08-04
fix(mcp): drop registryBaseUrl and index-url runtimeArguments from server.json

sgathrid:
Fixes the `@mcp` VS Code install of `io.github.Krv-Labs/topos`, which currently cannot boot:

```
error: the argument '--index-url <INDEX_URL>' cannot be used multiple times
Process exited with code 2
```

## Root cause — two independent failures

Both fields in the package record are wrong, and fixing **either one alone still leaves the install broken**.

**1. VS Code injects `--index-url` from `registryBaseUrl`, unconditionally.** `mcpManagementService.ts:136-141` (`microsoft/vscode@219ad590fe97`):

```ts
case RegistryType.PYTHON:
    if (serverPackage.registryBaseUrl) {
        args.push('--index-url', serverPackage.registryBaseUrl);
    }
```

A bare truthy check — no default-registry constant, no skip logic anywhere in the repo. So the `--index-url https://pypi.org/simple` added in #252 did not *add* the flag, it **duplicated** it, and `uv` rejects a repeated `--index-url` at argument-parse time.

**2. The only publishable base URL is not a usable index.** This is what kills a "just drop `runtimeArguments`" fix:

```
$ uvx --index-url https://pypi.org topos-mcp@0.4.3
  × No solution found ... topos-mcp was not found in the package registry
```

`https://pypi.org/topos-mcp/` returns 404 with **no redirect** to `/simple/`. Per PEP 503 the index base is `https://pypi.org/simple/`; the bare host is the human web UI. And `/simple` cannot be published — `internal/validators/registries/pypi.go:28-53` compares `registryBaseUrl != "https://pypi.org"` by exact string, no normalization.

## Why omitting the field is the fix

`registryBaseUrl` is **optional** — the `2025-12-11` schema's `Package.required` is `["registryType","identifier","transport"]`, with no `enum` or `pattern` on `registryBaseUrl`.

And the registry's default-fill never reaches the stored record: `ValidatePyPI` takes `pkg model.Package` **by value**, so `pkg.RegistryBaseURL = model.RegistryURLPyPI` mutates a local copy used only to build the ownership-check URL. Corroborated live — published PyPI servers `ai.adeu/adeu` and `ai.anomalyarmor/armor-mcp` serve no `registryBaseUrl` at all.

Resulting VS Code config, traced through every `args.push` site:

```json
"io.github.Krv-Labs/topos": { "type": "stdio", "command": "uvx", "args": ["topos-mcp@0.4.3"] }
```

Verified end-to-end with a real MCP stdio handshake:

```
$ echo '{"jsonrpc":"2.0","id":1,"method":"initialize",...}' | uvx topos-mcp@0.4.3
{"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2025-06-18",
 "serverInfo":{"name":"topos_mcp","version":"0.4.3+build.1785786694"}, ...}}
```

`runtimeHint: "uvx"` is kept for other clients, but note it is **not** what VS Code reads — `command` comes from a hardcoded `getCommandName(registryType)` map (`:183-191`), so **`registryType: "pypi"` is the load-bearing field**.

## Also in this PR

`extensions/vscode/workflow/mcp-registry-publishing.md` documented the unpublishable `registryBaseUrl: "https://pypi.org/simple"` at line 90 — almost certainly the origin of this loop. Corrected, with a callout recording all three traps.

Full investigation at `docs/reports/mcp-server-json-index-url.md`.

## Reviewer notes

- **`mcp-publisher validate` is not a preflight.** It passes all three shapes, including the unpublishable `/simple` one — `validate.go:28` calls only the schema validator; the base-URL guard lives in the publish path.
- **This needs the release to take effect.** `0.4.3` is published and `isLatest`; registry versions are immutable (`registry_service.go:188-194`) and GitHub-auth publishers get `publish` permission only, so the live broken record cannot be replaced in place. Order is PyPI → registry, and `release.yml` has no `mcp-publisher` step — the registry publish is manual.
- **`api.mcp.github.com` is currently serving degraded records** (drops `registry_type`, sends `registry_name: ""` → `convertRegistryType` falls through to `NODE` → `npx`). Mirror-wide, not topos-specific. If a post-release install emits `npx`, check the served `registry_type` before concluding this fix failed. The change is strictly non-worsening on that path too.
- **Not empirically proven:** that omitting `registryBaseUrl` publishes successfully. It is code-derived plus the two live examples above — proving it directly costs an immutable version number.

`scripts/check_versions.py` passes.


jeremy-wayland (changes requested):
The diagnosis holds and the fix is the right shape — deleting both fields is the only configuration that yields a working `uvx topos-mcp@<version>`, and `registryType: "pypi"` is correctly retained as the load-bearing field. The tracing in the description is thorough.

One blocking item, plus enforcement I've stacked as #277.

## Blocking

**The PR body points at a file this branch deletes.** It cites `docs/reports/mcp-server-json-index-url.md` for the full investigation, but `ef4e1de` ("docs: remove MCP index URL investigation report") removes those 288 lines. Anyone following the pointer after merge finds nothing. Please drop the sentence from the description — the callout added to `mcp-registry-publishing.md` already carries every load-bearing finding, so removing the report was the right call; the body just needs to catch up.

## Stacked: #277

Two things I'd rather not have this PR merge without:

1. **Nothing enforces the invariant.** These fields have now been added twice (#252, and the doc example this PR corrects). The answer here is a doc callout — but documentation did not prevent recurrence #2, and each recurrence burns an *immutable* registry version, since `0.4.3`'s record can never be replaced in place. Your own finding that `mcp-publisher validate` passes all three shapes means schema validation can never catch this: a repo-side guard is the only place it is catchable before publish. #277 extends the loop `check_versions.py` already runs over `.mcp/server.json` packages — no new file, no new CI step.

2. **This PR runs no CI.** `ci.yml:5-7` filtered `pull_request.branches` to `[main, worktree-rust-migration-v0.4.0]`, and this targets `release/044-256-harness-install` — so fmt, clippy, `cargo test`, `check_versions.py`, and the MCP stdio smoke test all skipped. Same for #271 and #275. #277 widens the filter to `release/**` rather than retargeting this one PR, which would have left the two siblings unchecked.

## Non-blocking

- `.mcp/server.json` still reads `0.4.3`, so the fix stays inert until the version bumps *and* someone runs `mcp-publisher publish` by hand — `release.yml` has no such step. Your reviewer notes call this out; flagging so it lands explicitly on the 0.4.4 release checklist rather than living only in a merged PR description.
- The doc example at `mcp-registry-publishing.md:81,92` still shows `0.4.1`. Pre-existing and not scanned by `check_versions.py`; a `<version>` placeholder would stop it going stale again, but leaving it is fine.
- Marking "omitting `registryBaseUrl` publishes successfully" as not empirically proven is the right disclosure. Given the by-value `ValidatePyPI` copy plus the two live PyPI servers that serve no `registryBaseUrl`, the code-derived argument is strong enough — proving it directly is not worth an immutable version number.

PR #278, opened 2026-08-04
Harness install: register a resolvable MCP server in every agent harness

jeremy-wayland:
Rewrites `topos install` / `uninstall` / `status` so every supported harness gets an MCP registration that actually resolves, and so uninstall leaves no trace.

Closes #285.

## Why

The draft reported **7/7 harnesses configured while only 3 worked**.

- **Copilot and Antigravity registered no MCP server at all.** Copilot got a prose block in `copilot-instructions.md`; Antigravity got a `GEMINI.md` `@import`. Neither file is where those tools look for MCP servers.
- **`{"command": "topos"}` cannot spawn from a GUI app.** Finder-launched macOS apps inherit only `/usr/bin:/bin:/usr/sbin:/sbin`, so Claude Desktop, Cursor and Antigravity failed on every install channel.
- **Uninstall left 16 directories behind** in a bare `$HOME`.

## What changed

Each harness now owns exactly one artifact — its MCP entry — written with an absolute command. Copilot moves to `~/.copilot/mcp-config.json`, Antigravity to `~/.gemini/config/mcp_config.json`, and the combined "Cursor & VS Code" target splits into two harnesses, VS Code taking `servers.topos` with `"type": "stdio"` in JSONC.

**Governing invariant: topos mutates only files it registers an MCP server into.** Prose blocks, `@import` lines and skill files are reported by `residue` and never touched — they are shared with other tools or owned by ClawHub/openclaw, and the harness skill directories are symlink farms whose deletion would reach into openclaw's namespace.

Behavior fixes, each measured:

- **Paths are never canonicalized before being recorded.** `/opt/homebrew/bin/topos` is a symlink into the Cellar, so pinning it breaks on `brew upgrade`. A PATH-alias search prefers the stable spelling. Verified against a Homebrew-shaped fixture: invoked *through the Cellar path*, it still records `bin/topos`, and the entry survives a simulated upgrade.
- **Drift is file identity, not string comparison** — one file under two spellings no longer rewrites configs on every run.
- **Entries compare field-wise** on `command`/`args`, preserving keys clients add instead of reporting them stale forever.
- **Backups are taken only when topos wrote nothing before**, so repairing drift cannot overwrite the pristine pre-install snapshot.
- **Writes follow symlinks**, so a config stowed into a dotfile repo stays a symlink.
- **Uninstall removes the directories install actually created**, tracked per write, only while empty.
- **`preserve_order`** stops install alphabetizing every key in `~/.claude.json`.

Adds a four-state model (`Active` / `Incomplete` / `Conflict` / `Absent`) that tells the user to re-run install rather than uninstall, a top-level `topos status` alongside the existing alias, and a three-way interactivity gate so `topos uninstall 2>log.txt` reports and exits non-zero instead of applying with no preview.

`integrations.rs` (740 lines, cyclomatic 140) is replaced by focused modules.

## Reconciling the parallel hardening pass

This branch picked up a competing rewrite of the same feature from the remote (`edc1011`, `989c9c2`). It converged independently on a harness table, but did not fix the defects above: Copilot and Antigravity still registered nothing, skill files were still written into `~/.agents`, no VS Code harness existed, directories were untracked, and its `current_exe().canonicalize()` is the Cellar-pinning bug.

Resolved in favor of this implementation, which fixes all of it and carries far more coverage (92 tests plus an end-to-end suite, versus 21 and none). Two additive pieces were **ported** from that pass: install now reports `repaired — ...` rather than restating the end state, and the per-module test scratch helpers collapse into one `testing::tmp_dir` whose atomic counter prevents same-label collisions under the threaded runner.

Everything outside the install module was taken from the remote unchanged: the `server.json` index-url fix, the CI invariant, the docs corrections, and the rustfmt fix.

## Also here

- **`fix(engine)`: upgrade tree-sitter-rust 0.23 → 0.24** (closes #285). A local named `raw` made a whole Rust file unparseable — topos scored it SLOP with no explanation. This shipped in our own tree: `topos/mcp/src/docs.rs` was silently unscoreable. Fixed upstream, so it is a version bump, not a workaround. Measured across all 171 Rust files here: one becomes parseable, **zero scores or raw metrics change anywhere else**. Regression tests cover both sides of the collision.
- **`ci`: give the `rust` job liblbug, and key the cargo cache to it.** The job runs clippy and the test suite, both of which link topos-engine, but only `composable` installed it — the failing check on this PR. See "Supersedes #284" below for why the missing step alone was not enough.

## Verification

- 560 tests across the workspace, 0 clippy warnings, rustfmt clean.
- End-to-end suite drives the real binary against a scratch `$HOME`. Both leave-no-trace assertions were mutation-checked to confirm they fail when they should.
- Manual: install → re-install → drift → repair → uninstall on a scratch home. Seeded foreign content survives byte-for-byte, and `numStartups` still precedes `mcpServers`, proving `preserve_order`.

## Not covered

Confirming the tools actually appear in **Claude Desktop, Cursor and Antigravity** needs a human looking at running apps — those are the three that cannot work today, so it is the check that matters most.

Four directories survive a *truly* bare `$HOME` (`~/.local`, `~/.local/state`, `~/Library`, `~/Library/Application Support`). They are in the `never_prune` set deliberately, since removing shared XDG/OS directories could break unrelated tools. Down from 16.

## Supersedes #284

#284 ("ci: unbreak the rust job") is closed in favor of this PR, which carries all of its content plus the fix it was missing.

| #284 change | Here |
| --- | --- |
| rustfmt `topos/engine/src/graphs/graphify/object.rs` | byte-identical (arrived via the remote merge, `3a9fe94`) |
| rustfmt `topos/mcp/src/schemas.rs` | byte-identical (same) |
| `ci.yml`: apt build deps + lbug cache + lbug setup in the `rust` job | present, plus the ordering and cache-key fix below |

**Why the extra fix was needed.** #284's diagnosis was right — the `rust` job runs clippy and the test suite, both of which link `topos-engine`, but only `composable` installed liblbug. Its remedy was incomplete: lbug's build script emits the `-L` search path only when `LBUG_LIBRARY_DIR` is set, and registers no `rerun-if-env-changed` for that variable. A `target/` restored from a run predating the variable therefore keeps stale build-script output and still fails to link, whatever the environment now says.

That went unnoticed because **the failure only appears on a cold cargo cache, and no recent `rust` run had one.** #284's own green tick came from a primary-key cache hit where lbug was never recompiled — as did every prior green run. Changing `Cargo.lock` here (the tree-sitter bump) dropped the job to a restore-key fallback and surfaced it, which is why the first push of this branch still failed with `could not find native static library lbug` *after* the #284 steps were applied.

The fix moves the lbug setup ahead of the cargo cache so `LBUG_LIBRARY_DIR` is exported before `rust-cache` computes its key, and adds that variable to `env-vars` so a cache built without lbug is never restored into a job that needs it.

**Verified cold, not cached.** The passing run compiled lbug three times from a fully cold cache (351 crates built from scratch), so this is the first time the lbug path has actually been exercised in the `rust` job rather than restored around.

## Breaking

The `skills` harness id is replaced by `cursor` and `vscode`, and `topos install` no longer writes skill files or instruction blocks.


jeremy-wayland:
The `rust` check on this PR is blocked by two pre-existing failures on `release/v0.4.4`, not by anything in this branch. Both are fixed in #284:

1. **`Format check`** — rustfmt drift in `engine/graphify/object.rs` and `mcp/schemas.rs` (unpinned `@stable` rustfmt; files this PR does not touch). This branch carries the same fix, so its format check already passes.
2. **`Clippy`** — `could not find native static library \`lbug\``. The `rust` job never sets up liblbug, while `composable` in the same `ci.yml` does. Invisible until now because the format check failed first and clippy never ran.

#284 is green on all of `Format check`, `Clippy`, `Tests`, and the MCP smoke test — the first time those have run on this branch family. Once it merges, rebase this branch and the `rust` check should pass on the harness work itself.

Nothing to change here in the meantime.

PR #279, opened 2026-08-04
release: v0.5.0

### LINKED ISSUE #66: Automate MCP Registry validation and publishing
## Context

Topos is now published to the official MCP Registry as `io.github.Krv-Labs/topos`, but registry publishing is still manual:

```bash
mcp-publisher validate .mcp/server.json
```

The MCP Registry docs now recommend automating registry publication from GitHub Actions using OIDC (`mcp-publisher login github-oidc`) and note that no dedicated MCP secret is required for GitHub OIDC auth. See: https://modelcontextprotocol.io/registry/github-actions && docs here: https://github.com/modelcontextprotocol/registry/blob/main/docs/modelcontextprotocol-io/github-actions.mdx

## Why

Manual registry publishing creates a release gap after PyPI succeeds. The registry validates that the package exists in the underlying package registry, so the MCP publish step should run after the PyPI publish job has completed successfully.

## Proposed implementation

Add a GitHub Actions job, likely in `.github/workflows/release.yml`, that:

- Runs only for release publishing events (`push` tag and appropriate `workflow_dispatch` paths).
- Depends on `publish-pypi` so `topos-mcp==<version>` exists before MCP Registry validation.
- Uses `permissions:
  id-token: write
  contents: read` for MCP Registry OIDC authentication.
- Checks out the repo.
- Installs `mcp-publisher` from the official registry release artifact.
- Authenticates with:

```bash
./mcp-publisher login github-oidc
```

- Validates the checked-in metadata:

```bash
./mcp-publisher validate .mcp/server.json
```

- Publishes the checked-in metadata:

```bash
./mcp-publisher publish .mcp/server.json
```

## Secrets / environment

- Preferred: GitHub OIDC auth, which should not require a dedicated MCP secret.
- Confirm GitHub OIDC works for the `io.github.Krv-Labs/topos` namespace from this org repository.
- If OIDC cannot satisfy org namespace permissions, document and add a fallback `MCP_GITHUB_TOKEN` secret with the registry-documented `read:org` and `read:user` scopes.
- Keep PyPI publishing as the source package gate; MCP registry publish should not run before PyPI succeeds.

## Acceptance criteria

- `mcp-publisher validate .mcp/server.json` runs in CI before publishing.
- `mcp-publisher publish .mcp/server.json` runs automatically after successful PyPI publish.
- The workflow uses GitHub OIDC unless proven incompatible with the org namespace.
- The job is skipped for pull requests.
- The job does not republish VS Code Marketplace or GitHub release artifacts.
- The issue/PR notes any required repo/environment secret changes.

## Notes

Current server metadata is already checked in at `.mcp/server.json`; this issue is only for automation and release sequencing.

### LINKED ISSUE #210: Parity: C++ declaration-only nodes map to VarDecl, not FunctionDecl
_From the review of #159 (Rust v0.4.0 migration). Non-blocking._

**Where:** `topos/engine/src/graphs/uast/mapper_cpp.rs` — `map_node_kind` / `is_declaration_only_kind`

**What:** Python's cpp mapper inspects the node (`_has_function_declarator`): a declaration-only signature (`void f(int);`, or pure-virtual `virtual double area() const = 0;`) → `FunctionDecl`; a plain variable/member → `VarDecl`. The Rust `map_node_kind` callback only receives the *kind string* (not the `Node`), so `is_declaration_only_kind` unconditionally returns `VarDecl`.

**Impact:** every header-only prototype and pure-virtual method signature that Python emits as `FunctionDecl` is emitted as `VarDecl` by Rust → SIMPLE structural node-kind counts diverge for typical C++ headers and abstract base classes. Already acknowledged with a `ponytail:` comment, but it's a real divergence.

**Fix:** widen the `map_node_kind` callback to take `&Node` so the declaration-only branch can replicate `has_function_declarator` (the node-aware `has_function_declarator` already exists in this file for the abstractness path).

### LINKED ISSUE #211: GitNexus fingerprint written to flat .gitnexus, not branch-scoped store
_From the review of #159 (Rust v0.4.0 migration). Non-blocking._

**Where:** `topos/engine/src/adapters/gitnexus.rs` — `finished_result`

**What:** always writes `.topos-fingerprint.json` into `target_dir/.gitnexus`. Python resolved the active store first (`current_git_branch` → `resolve_lbug_store`) and wrote the marker next to the branch-scoped store (`.gitnexus/branches/<branch>-<hash>/`) when GitNexus wrote there.

**Impact:** dormant today — no code reads `.topos-fingerprint.json` back (grep finds only the writer + its tests). But any future staleness reader keyed on the resolved store dir would never find it in a branch-scoped repo and would regenerate unconditionally.

**Fix:** mirror Python — resolve the store via `current_git_branch` + `resolve_lbug_store` and write the fingerprint into the resolved store's parent, falling back to the flat `.gitnexus`.

### LINKED ISSUE #212: Parity: LadybugDB loader ingests only File nodes; drops edge confidence/reason
_From the review of #159 (Rust v0.4.0 migration). Non-blocking. Related: #198._

**Where:** `topos/engine/src/graphs/mdg/ladybug.rs` — `from_ladybug_native`

**What:** the native loader only queries `MATCH (n:File)` and stubs every other referenced id as a bare `Symbol` node with empty properties; the relationship query never selects `confidence`/`reason`, hard-coding `confidence: 1.0` / `reason: ""`. Python's `_from_ladybgdb` discovered all node tables (`CALL show_tables()`), loaded every node (Class/Function/Method/Import/…) with its real label and properties, and selected `r.confidence`/`r.reason` per edge.

**Impact:** today's MDG metrics read only the `File` label + `filePath` and traverse edges by id/type, so current outputs are unaffected — but any metric/consumer inspecting non-File labels, node properties, or edge confidence/reason silently diverges from Python.

**Fix:** either document this as a permanent narrowing (confidence/reason + non-File metadata not consumed) or restore full node-table discovery + confidence/reason selection.

### LINKED ISSUE #225: CFG: model switch breaks and try/catch flow before expanding golden coverage
## Context

PR #224 adds a compact CFG edge-contract matrix: 16 cleanly parsed cases across all six supported languages, baselined against the recursive builder at `118767b`. It covers branch/loop, match/switch, and the languages that map `TryStmt` without snapshotting raw block IDs.

The matrix intentionally preserves existing behavior. Before expanding it into exhaustive goldens, we should resolve two ambiguous CFG semantics so tests do not make accidental behavior permanent.

### Switch `break` targets

The builder documentation says `break` resolves against the innermost loop or switch, but only loop construction pushes a `LoopContext`. A JavaScript/C++ switch nested inside a loop can therefore resolve `break` against the enclosing loop rather than the switch join.

Define and test these cases explicitly:

- `break` exits the innermost switch to its join block
- `continue` inside a switch nested in a loop continues the loop
- nested loop/switch combinations select the correct target

### Try/catch/finally flow

The current `TryStmt` model emits one coarse exception edge to `try_join` and does not model handlers or `finally` separately. In the new baseline contract, a source where both the try body and catch handler return contains only one modeled `ReturnStmt`.

Decide the intended structural contract for:

- individual catch/except handlers
- handler returns and throws
- `finally` fallthrough/termination
- multiple handlers

## Follow-up coverage

After those semantics are settled, extend the stable edge contracts with focused cases for:

- switch `break` and nested loop/switch targets
- try/catch/finally handler flow
- Go discriminant-less and type switches
- representative nested-if mapper shapes
- TypeScript-specific syntax where its tree shape differs from JavaScript

Keep expectations semantic: normalized block shapes and edge labels, no raw block IDs. Every fixture should assert a clean parse, and the supported-language meta-guard should remain fail-closed.

## Acceptance criteria

- Switch and try semantics are documented in the builder.
- Focused tests fail on the current incorrect/underspecified behavior and pass after the implementation.
- Expanded contracts cover the resolved shapes without introducing snapshot dependencies or a many-file fixture corpus.
- `cargo test --workspace` passes.

### LINKED ISSUE #230: Go files get an empty module-level callable in the CFG, inflating counts
Found during review of #224 (stacked on #159). Backlog — deliberately **not** fixed in #224.

**Problem:** When the CFG builder collects module-level statements, Go's `package` clause survives the filter, so every Go file grows an extra, empty module-level callable alongside its real functions. The new cross-language edge-contract tests in #224 make this visible: Go fixtures carry one extra block and two extra edges compared to the identical program in every other language.

**Result:** Go CFGs report slightly inflated block/edge counts, which nudges cyclomatic complexity one higher per file than the equivalent code in other languages — a small cross-language parity skew in SIMPLE scoring.

**Why not fixed in #224:** the edge-contract tests intentionally lock the pre-refactor builder's behavior to prove the iterative rewrite changed nothing. Fixing this parity skew is a real behavior change and belongs in its own PR, updating the Go contracts alongside it.

**Proposed fix:** exclude package/import clauses when collecting module-level statements, then update the locked Go edge contracts to the corrected shapes.

### LINKED ISSUE #231: cfg.nesting_depth doesn't restore depth at branch joins — reports 1008 for a 460-line file
## What

`cfg.nesting_depth` is documented as "Max static nesting depth" but does not measure nesting. It reports **1008** for `topos/engine/src/adapters/discovery.rs` (460 lines) and **1134** for `topos/mcp/src/formatting.rs` (846 lines).

## Why

`ControlFlowGraph::max_nesting_depth` (`topos/engine/src/graphs/cfg/object.rs:96`) relaxes a depth value along edges, incrementing on `EdgeKind::True | SwitchCase` and taking the max at each node. Depth is never **restored at a branch join** — the join inherits the incremented value — so sequential, non-nested branches accumulate. On the merged whole-file CFG this sums across every function in the file.

Small fixtures look correct only because they have too few branches to expose it:

| Fixture | Reported | True max nesting |
|---|---:|---:|
| 3 nested `if`s (py / rs) | 3 | 3 |
| 1 flat `if` | 1 | 1 |
| `discovery.rs` (460 lines) | **1008** | ~3 |

The test at `object.rs:288` is named `max_nesting_depth_of_sequential_true_branches` but actually builds a *nested* chain (`0 -True-> 1 -True-> 2`), so it does not pin the sibling-accumulation behavior.

## Impact

Diagnostic-only — `cfg.nesting_depth` is not in `GATE_SPECS`, so **no medal, score, or verdict is affected**. But the value is surfaced in `raw_metrics` to agents and in `topos inspect`, where it is meaningless at this magnitude.

## Fix sketch

Depth must fall back at the post-dominator of a branch. Needs post-dominator-aware traversal rather than the current monotonic relaxation. Note this touches the CFG that the v0.4.0 stack rewrote under golden edge contracts (`cfg/edge_contracts.rs`) — those lock edge *shapes*, not this derived metric, but the change deserves its own PR and fixture review.

Found while verifying the MCP interface pass on the #159 stack.

### LINKED ISSUE #233: agent_contract is built by four independent builders that duplicate flag logic
## What

`AgentContract` is constructed in four separate places that each re-implement the same non-COMPOSABLE flag logic, plus inline literals in a fifth:

| Builder | Location |
|---|---|
| `build_agent_contract` | `topos/mcp/src/formatting.rs` (evaluate_code / evaluate_file / inspect_code) |
| `project_contract` | `topos/mcp/src/tools/evaluate.rs` |
| `assessment_contract` | `topos/mcp/src/tools/assess.rs` |
| `changeset_contract` | `topos/mcp/src/tools/assess.rs` |
| inline literals | `topos/mcp/src/tools/depgraph.rs` + every error path |

Only `composable_contract_signals` is shared. Each builder separately re-implements `if !warnings.is_empty() { risk_flags.push("warnings") }` and its own `grade_capped` / `active_security_findings` pushes, and each carries its own hardcoded `verification_gates` vector.

## Observed divergence

`assessment_contract` passes `include_missing = false` while the other three pass `true`, so `missing_gitnexus_dir` never appears in an assessment's `blocked_by`. That asymmetry is undocumented and looks unintentional.

## Impact

No current user-visible bug beyond the above, but every contract change has to be made four times, and drift between them is invisible until dogfooded. Several fixes in the #159 interface pass had to be applied to more than one builder for this reason.

## Related

`blocked_by` / `risk_flags` are bare `Vec<String>` with no enum, and `composable_contract_signals` recovers codes by **substring-matching marker constants against prose warning text** — so any warning-text edit can silently change `blocked_by`. Worth considering structured codes (`code=…`) at the same time. See also the marker-survival tests added in `topos/mcp/src/evaluation/mod.rs`.

### LINKED ISSUE #234: Project agent_contract routes to inspect_code, not evaluate_file on the ROI head
## What

After `topos_evaluate_project`, `agent_contract.next_tool` is `topos_inspect_code`. For "find the cheapest structural win in this repo", the useful next call is `topos_evaluate_file(refactor_targets=…)` on a ranked head, then inspect only if that is insufficient.

Compounding it, `worst_files` ranks by ascending minimum score across dimensions. On a repo with many leaf modules at `mdg.instability = 0.0`, COMPOSABLE zeros flood the head of the list, so the "start here" pick is often a file that is already `SIMPLE_SECURE` with a healthy simple score — while genuine SIMPLE gate failures sit further down the page.

## Evidence

Reported independently twice: once in an agent post-mortem ("worst file ≠ lowest-hanging fruit"; the guidance pointed at `cli/src/commands/compare.rs` for composable weakness while the real SIMPLE fails were `file_roles.rs` at max_fn 30 and `gitnexus.rs`), and again in a later dogfood of the rebuilt binary, which found the floor at SLOP with the page mostly `SIMPLE_SECURE`.

## Why it isn't just a `next_tool` swap

The underlying request is a ranking mode. Options worth weighing:
- `rank_by = roi | max_function_complexity | simple_gap | composable_gap`
- or separate lists: `hard_fails[]` vs `leaf_composable_zeros[]` vs `maintainability_giants[]` (high cyclomatic but already passing)

Either way it needs a definition of ROI that survives review, which is why it was scoped out of the MCP interface pass on the #159 stack rather than guessed at.

## Note

`worst_files` is now documented as page-global by the existing sort, explicitly **not** ROI, so the current behavior is at least no longer mis-advertised.

### LINKED ISSUE #282: NAVIGABLE thresholds are PROVISIONAL — calibrate before tagging v0.5.0
## What

`NAVIGABLE`'s gate thresholds are **PROVISIONAL** and must be replaced with calibrated values before `v0.5.0` is tagged. `calibration.rs` says so in the type's own doc comment:

```rust
pub const NAVIGABLE: NavigablePolicyThresholds = NavigablePolicyThresholds {
    max_function_divergence: 8.0,
    divergence_cap: 20.0,
};
```

> PROVISIONAL: a first-pass estimate, not yet run through the corpus ECDF calibration the SIMPLE and SECURE constants received. […] Must be replaced with a calibrated value before v0.5.0 is tagged; see the release checklist.

Filed so this does not get discovered at tag time. Introduced by #280.

## Where

`topos/engine/src/evaluation/policies/calibration.rs` — `NAVIGABLE`

## Current basis

`8.0` was reasoned from structure, not from a corpus: roughly "a function may nest about four levels deep before an agent starts paying for it" — a 4-deep single-child chain scores `4.16`, and realistic nesting with fan-out reaches the bound around the same depth.

Measured against Topos's own 157 Rust files, Semantic Compositional Divergence distributes as:

| Statistic | Value |
| --- | --- |
| median | 0.69 |
| p75 | 2.08 |
| p90 | 4.56 |
| p95 | 5.98 |
| **8.0 lands near** | **p97** |
| files failing | 2.5% |

`divergence_cap: 20.0` (score normalization, never gates `achieved`) has no empirical basis at all.

## The decision this needs

**What failure rate is the gate supposed to produce?** Every other pillar's gate was set from a target percentile against a corpus, not from a target failure count. `8.0` sitting at p97 is a coincidence of the structural reasoning, not a chosen operating point.

Sub-questions:

1. **Which corpus?** Topos's own 157 Rust files are a single-language, single-style sample. SIMPLE and SECURE were calibrated more broadly — NAVIGABLE gating on a Rust-shaped distribution risks mis-gating Python and Go, whose idiomatic nesting differs (Go's `if err != nil` fan-out, Python's lack of braces encouraging flatter code).
2. **Per-language thresholds, or one?** The other pillars use one. If the distributions differ materially across the six supported languages, one threshold means the gate is stricter for some languages than others — the same cross-language parity concern as #230.
3. **What should `divergence_cap` be?** It only shapes the reported score, but a cap of `20.0` against a p95 of `5.98` means nearly every real file scores in the top quarter of the curve, so the score carries little discriminating signal.

## Done when

- `max_function_divergence` and `divergence_cap` are set from a stated percentile against a named corpus, with the methodology recorded next to the constants.
- The `PROVISIONAL` doc comment is removed or replaced with the calibration basis.
- Any golden/fixture expectations that moved are updated.

## Related

- #280 (introduces NAVIGABLE), #279 (v0.5.0 release checklist), `docs/decisions/v0.5.0-stack.md` in #281
- #230 — the cross-language parity concern in sub-question 2 has precedent

jeremy-wayland:
Integration PR for **v0.5.0**. Child PRs merge into `release/v0.5.0`; this PR
squash-merges into `main` at release time.

## Release summary

v0.5.0 adds NAVIGABLE as a fourth independent quality pillar and tightens the
meaning of file-level COMPOSABLE.

- Ω grows from 8 to 16 elements; `LatticeElement` gains 8 MCP wire variants.
- `IDEAL` now requires SIMPLE, COMPOSABLE, SECURE, and NAVIGABLE.
- Medal tiers band over four satisfied pillars: PLATINUM / GOLD / SILVER /
  BRONZE / SLOP.
- Rankings must contain all four pillars. The default is
  `SIMPLE ≻ NAVIGABLE ≻ SECURE ≻ COMPOSABLE`.
- NAVIGABLE gates per-function divergence at `10.0`, with a `12.0` score cap,
  calibrated on a balanced 6,390-file leaderboard corpus.
- File COMPOSABLE gates only `mdg.fan_out <= 10`, calibrated on 2,979
  production files across Python, Rust, TypeScript, and MCP cohorts. Fan-in and
  Martin stability readings remain scored/actionable advisories; coupling,
  abstractness, and dependency depth remain inspect diagnostics.

This is a pre-1.0 breaking minor release. The full release record is in
`CHANGELOG.md`, dated 2026-08-07.

## Issues closed by this release

Child PRs target the release branch, so closing keywords are declared here for
the eventual merge into the default branch.

Closes #66
Closes #210
Closes #211
Closes #212
Closes #225
Closes #230
Closes #231
Closes #233
Closes #234
Closes #282
Closes #283
Closes #287
Closes #288
Closes #289

| Issue | Resolution |
| --- | --- |
| #282 NAVIGABLE thresholds | Gate `10.0`, cap `12.0`, balanced 6,390-file calibration |
| #231 / #288 CFG nesting | Post-dominator region walk restores joins and ignores loop back edges |
| #230 Go phantom callable | Module preambles no longer create implicit callables |
| #210 C++ declarations | Header prototypes and pure-virtual declarations map as functions |
| #211 GitNexus fingerprint | Stored beside the resolved branch-scoped graph |
| #212 LadybugDB ingestion | All discovered node labels and relation evidence are loaded |
| #225 switch / try / finally | Correct break, handler, and nested-finally completion semantics |
| #233 AgentContract drift | Shared composability/parse/security/grade-cap prelude |
| #234 project routing | Gate-first `hard_fails`, deprecated `leaf_composable_zeros`, and maintainability lists |
| #66 MCP Registry publishing | OIDC publisher is pinned and checksum-verified |
| #287 missing graph override | In-root missing stores can generate; warnings remain explicit |
| #289 named-file inference | Supported suffixes infer language without `--language` |
| #283 lattice figures | Both themes show the sixteen-element four-pillar lattice |

## Landed stack

- #280 — NAVIGABLE and the sixteen-element Ω
- #294 — CFG metric correctness
- #295 — C++ declaration and branch-fingerprint parity
- #296 — full LadybugDB ingestion
- #297 — MCP Registry publishing
- #298 — shared AgentContract prelude
- #299 / #304 — switch/try/finally correctness and NAVIGABLE walk metadata
- #300 — gate-first project ranking lists
- #301 / #302 — calibration, four-pillar docs, preferences, and wire invariants
- #303 — release records and publishing hardening
- #308 — VS Code dependency patches
- #309 — per-call MCP project-root derivation
- #312 — instability resolution-limit correction
- #313 — calibrated file-level fan-out gate and retained composability diagnostics

## Explicitly deferred

- #286 — improve `cfg.cyclomatic` remediation wording.
- #293 — verify MCP gallery workspace binding against the published v0.5.0.
- #305 — precompute project ranking sort keys.
- #306 — reuse parsed morphisms for metric locations.
- #307 — simplify residual CFG context wrappers.
- Package-level stability remains a separately scoped post-v0.5 redesign; it
  must not be inherited into every member file.

## Verification

PR #313 merged green (`rust`, `composable`, `wheel`, `extension`). The merged
release branch is also verified locally with:

- `cargo test --workspace`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo fmt --all --check`
- `python3 scripts/check_versions.py`
- `python3 scripts/check_skill.py`
- `python3 scripts/generate_self_badge.py --self-check`

## Review focus for @sgathrid

1. The four-pillar wire and medal changes are intentional for the minor bump.
2. COMPOSABLE's file verdict is narrowly fan-out; richer findings remain in
   score/inspect/refactor rather than being discarded.
3. The published thresholds, README, embedded MCP docs, Sphinx docs, decision
   records, and changelog now agree.
4. Package-level stability is explicitly deferred instead of projected onto
   files.

## Release checklist

- [x] v0.4.4 merged and `main` synchronized
- [x] NAVIGABLE calibrated at gate `10.0` / cap `12.0`
- [x] COMPOSABLE calibrated at file fan-out `10`
- [x] four-generator agent/MCP documentation and lattice figures updated
- [x] v0.5.0 changelog dated and current
- [x] #312 and #313 merged into `release/v0.5.0`
- [x] local release validation passes
- [x] refreshed PR #279 CI is green
- [ ] final review from @sgathrid


jeremy-wayland:
@sgathrid can I pass you the token to regenerate the lattice figures now for a lattice generated by 4 elements?

jeremy-wayland:
## Lattice figures redrawn for four generators (f0c9d73)

`docs/source/_static/figures/topos-lattice{,-dark}.svg` now show the full sixteen-element 4-cube — `SLOP` → four BRONZE singles → six SILVER pairs → four GOLD triples → `IDEAL` (🏆 PLATINUM) — instead of the three-generator sub-cube. Pair and triple node fills are RGB blends of their generators (same tinting the old figure used for `S ∧ C`); per-edge arrowheads are gone, since 32 of them read as noise and the vertical guide already carries the direction of the order.

Also swept the prose that still described a three-pillar model: `index.rst` (alt + caption), `measures.rst` (8-valued/3 generators → 16-valued/4), `concepts.rst`, `installation.rst`, `quickstart.rst`, `cli.rst`, `agents.rst`, `architecture.rst`, the VS Code extension README/`package.json`/`extension.ts` (its medal table still topped out at GOLD = all 3), the crate and `pyproject` descriptions, and the engine/MCP doc comments. `cargo test -p topos-mcp --lib context_budget` still passes — the description edits fit inside the ratchet.

Left as history on purpose: `omega.rs`, `schemas.rs`, `context_budget.rs`, `config.rs`, `topos://docs/lattice`, and the `versionchanged:: 0.5.0` notes all describe the three-generator algebra as the *former* state, which is correct.

### @sgathrid — still needs regenerating for four pillars

These four figure pairs plot an actual benchmark run over `requests` / `numpy` / `pandas`, so they need the corpus **re-run under NAVIGABLE** rather than redrawn. I did not want to hand-edit real measurements:

| Figure | What is stale |
| --- | --- |
| `topos-medal-mix{,-dark}.svg` | Medal-mix bars are Gold/Silver/Bronze/Slop — no PLATINUM band, and the percentages come from three-pillar banding |
| `topos-typical-profiles{,-dark}.svg` | Median per-pillar scores for Simple/Secure/Composable only — no NAVIGABLE series |
| `topos-library-profiles{,-dark}.svg` | Same three-series bars per library — no NAVIGABLE series |
| `topos-bronze-ready{,-dark}.svg` | "files strong enough for bronze by pillar" over three pillars — no NAVIGABLE column |

They're referenced from `docs/source/measures.rst` and `docs/source/index.rst`. Tracked as the remaining half of #283, and noted in the CHANGELOG `Notes` section plus `docs/decisions/v0.5.0-stack.md` under deferred work.

The lattice SVGs were emitted from a small throwaway script rather than hand-authored — happy to land it under `scripts/` if we'd rather these be reproducible than hand-maintained.

PR #280, opened 2026-08-04
feat: NAVIGABLE — a fourth generator of Ω

jeremy-wayland:
**PR 1 of the v0.5.0 stack.** Base: `release/v0.5.0` (#279). Merge with a **merge commit**, not squash.

Adds `NAVIGABLE` as a fourth generator of $\Omega$: a file is navigable when its *worst* function is not too deeply nested.

## Why nesting, and why it isn't `SIMPLE`

Classical complexity metrics stop predicting LLM task accuracy once code length is controlled for; nesting depth keeps predicting it. A flat function with six sequential `if`s and a six-deep nested one carry the same cyclomatic complexity, but models reason reliably over the first and fail on the second — each nesting level is another hierarchical state attention has to hold open.

So the probe measures nesting and only nesting. For every scope-forming node $u$ inside a callable:

$$\mathrm{SCD}(f) = \sum_u \mathrm{depth}(u)\cdot\ln\!\left(1 + \mathrm{fanout}(u)\right)$$

Two properties of that shape are load-bearing:

- A leaf scope contributes $\ln 1 = 0$, so a perfectly flat function scores exactly `0.0`. Branch *count* stays SIMPLE's concern via `ast.max_function_complexity`; the weight lands on the ancestors doing the nesting.
- `ConditionalExpr` and short-circuit `BinaryExpr` are excluded — expression-level branching opens no block, so it costs no state to track, and counting it would re-measure SIMPLE.

The gate is the **per-function max**, not a file-wide sum. This repo already learned that once: `cfg.cyclomatic` is a whole-file merged-CFG sum that scales with function count, so it was demoted to advisory (#193) and `ast.max_function_complexity` gates in its place. A file-wide SCD sum would fail a long file of short flat functions for its length.

## Commits

| Commit | Scope |
| --- | --- |
| `92ae94e` | `feat(engine)`: Semantic Compositional Divergence probe (`functors/probes/ast/divergence.rs`, `scopes.rs`) |
| `40e483f` | `feat(engine)`: extend $\Omega$ from 8 to 16 elements with the `NAVIGABLE` generator |
| `0d16f2d` | `feat(engine)`: wire `NAVIGABLE` into $\chi_S$ with a gate and policy translator |
| `4ca779a` | `feat(mcp,cli)`: surface `NAVIGABLE` across the wire and terminal |
| `10fc4e1` | `docs`: describe `NAVIGABLE` and the 16-element lattice |

Follows the four-step "add a representation" recipe in `.agents/AGENTS.md`: probe → `Representation` metrics (`nav.*`) → `GATE_SPECS` registration → policy translator.

## Behavior changes reviewers should look at

- **$\Omega$: 8 → 16 elements.** Free Heyting algebra on 4 pairwise-incomparable generators; `LatticeElement` gains 8 wire variants.
- **`IDEAL` now requires all four pillars** — the former `IDEAL` is now `SIMPLE_COMPOSABLE_SECURE`. **CI pinned to `IDEAL` will start failing.** This is the change most likely to surprise a downstream consumer.
- **Medals re-grade in both directions**, banding on satisfied-pillar count: 4 → `PLATINUM`, 3 → `GOLD`, 2 → `SILVER`, 1 → `BRONZE`, 0 → `SLOP`. `GOLD` is not renamed; it now means 3 of 4, so the old top verdict lands there.
- **Preference ranking is 4 wide.** Rank weights go `4/2/1` → `8/4/2/1`; default ranking is `(SIMPLE, COMPOSABLE, SECURE, NAVIGABLE)`. A three-element `.topos.toml` ranking is no longer a permutation of `G_qual`, so it is dropped and the default applies — matching the best-effort contract other malformed keys get. Worth confirming that silent-drop is preferred over a warning.
- **MCP tool-definition context budget grows 33,736 → 39,584 chars** from the larger `LatticeElement` / `GeneratorInput` enums appearing in every schema that accepts a verdict or ranking.
- **Vacuous pass on unparseable input**, matching `Φ_SIMPLE` — no metric means `achieved = true`, score `1.0`.
- **Gate operations reuse the existing vocabulary** (`extract_helper`, `split_decision_logic`) rather than inventing NAVIGABLE-only verbs, since the fix for a divergence failure is the same as for a complexity failure.

## Known gap — blocks the tag, not this PR

`NAVIGABLE`'s thresholds are **PROVISIONAL** and marked as such in `calibration.rs`:

```rust
max_function_divergence: 8.0,
divergence_cap: 20.0,
```

They are first-pass estimates ("a function may nest ~4 levels before an agent pays for it"), not run through the corpus ECDF calibration that SIMPLE and SECURE constants received. Over Topos's own 157 Rust files the distribution is median `0.69`, p75 `2.08`, p90 `4.56`, p95 `5.98` — so `8.0` sits near p97 and fails 2.5% of files.

Worth settling **what failure rate the gate should produce** before picking the calibrated value: every other pillar's gate was set from a target percentile rather than a target failure count, so NAVIGABLE should be too. Tracked on #279's release checklist.

Also stale, and cheap to miss: `docs/source/_static/figures/topos-lattice*.svg` still depict the three-generator sub-cube. This PR makes the captions say so explicitly, but the artwork has not been redrawn for 16 elements.

## Verification

- `cargo test --workspace` — 484 passed, 0 failed (348 engine / 54 cli / 82 mcp)
- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets` — zero warnings
- `python scripts/check_skill.py` — passes

## Note on an earlier comment

My first review comment on this PR claimed it silently resolved #193 by making `cfg.cyclomatic` advisory. **That was wrong** — `gates_achieved: false` has been on `main` since v0.4.3 (#251), and this PR's `gates.rs` diff adds only `interpret_divergence` and the `nav.max_function_divergence` spec. Corrected in a follow-up comment; withdrawn as D1 in #281. There is no SIMPLE-semantics change here.

## Still open for your call

- **Threshold calibration methodology** — what failure rate should the gate target? Every other pillar's gate came from a target percentile; `8.0` was picked as "~4 levels of nesting" and lands near p97 by coincidence.
- **Silent-drop of three-element rankings** — a `.topos.toml` ranking that lists three pillars is dropped in favor of the default. Consistent with other malformed keys, but a warning may serve users better than silence during the upgrade.
- **Lattice figures** — `docs/source/_static/figures/topos-lattice*.svg` still show the three-generator sub-cube. Captions say so; artwork is not redrawn.


jeremy-wayland:
## This PR silently resolves #193

`gates.rs` in this branch sets `gates_achieved: false` on `cfg.cyclomatic`, with the rationale in a code comment citing #193:

```rust
metric: "cfg.cyclomatic",
pillar: "simple",
...
// Advisory only (issue #193): the whole-file merged-CFG sum
// scales with function count, so it shouldn't hard-fail a file
// for having many small, individually-simple functions.
// ast.max_function_complexity gates that concern directly.
gates_achieved: false,
```

That is **option 2 of the two options #193 lists**, and it is the only `gates_achieved: false` in `GATE_SPECS`. On `main`, `cfg.cyclomatic` is still a hard SIMPLE gate.

The change looks correct on the merits — #193's own evidence is that 33/99 `topos-engine` files fail SIMPLE on `cfg.cyclomatic` alone while passing `ast.max_function_complexity`, and that splitting `graphs/mdg/object.rs` by responsibility took cyclomatic 48 → 25 + 24, both still failing the flat threshold of 15. Gating per-function is the right call, and `divergence.rs`'s doc comment argues from exactly that principle, so NAVIGABLE would be internally inconsistent without it.

The concern is process, not correctness. #193 asked for the opposite, verbatim:

> Deliberately not fixed in PR #191 — changes SIMPLE's meaning for every future evaluation, wanted a separate decision/review rather than folding it into a cleanup pass.

Right now it is folded into a PR titled "NAVIGABLE", so a reviewer looking for a SIMPLE-semantics change won't find it by title, and it has no CHANGELOG entry. Every file in every repo that was failing SIMPLE on aggregate cyclomatic alone starts passing.

### Options

- **A** — split `gates_achieved: false` into its own chunk PR against `release/v0.5.0`, land it before this one, close #193 with that PR. One extra small PR; the SIMPLE change gets its own review and CHANGELOG entry.
- **B** — keep it here, but expand this PR body to say it resolves #193 and add a `### Changed` CHANGELOG entry for it.
- **C** — revert it and leave #193 open. Not recommended: leaves the codebase arguing for per-function gating in `divergence.rs` while gating on a file-wide sum.

Full write-up as **D1** in `docs/decisions/v0.5.0-stack.md` (#281).

## Other items from the review

- **`.agents/AGENTS.md` is not updated.** It still says "8-element lattice ($\Omega$)" and "3 independent, pairwise incomparable generators". This PR updates `docs/source/`, `topos/mcp/docs/content/`, and `skills/topos/SKILL.md`. `.agents/AGENTS.md` is the canonical agent rules file *and* one of the three paths OpenWiki's PR is allowed to touch, so a hand-edit here needs to survive regeneration.
- **Thresholds are `PROVISIONAL`** (`max_function_divergence = 8.0`, `divergence_cap = 20.0`) and `calibration.rs` says they must be replaced with corpus-ECDF-calibrated values before v0.5.0 is tagged. Tracked on #279's checklist — flagging so it isn't discovered at tag time.

## Verified locally

Rebased onto `release/v0.5.0` (clean; the one conflict was `SKILL.md`, where v0.4.4's `--gitnexus-dir` guidance and this branch's `GOLD` → `PLATINUM` rename both needed keeping — resolved by taking both).

- `cargo test --workspace` — 484 passed, 0 failed (348 engine / 54 cli / 82 mcp)
- `cargo fmt --all --check` — clean
- `cargo clippy --workspace --all-targets` — zero warnings
- `python scripts/check_skill.py` — passes

jeremy-wayland:
## Correction: disregard my #193 comment above — this PR does not touch `cfg.cyclomatic`

I got that wrong. `gates_achieved: false` on `cfg.cyclomatic` is **already on `main`** — it shipped in **v0.4.3 via #251**, not in this PR. The diff of `gates.rs` against `release/v0.5.0` adds only `interpret_divergence` and the `nav.max_function_divergence` gate spec; it leaves the `cfg.cyclomatic` entry untouched. This PR doesn't modify `simple.rs` at all.

`main`'s `simple.rs` module header already documents the resolved state:

```
//! Φ_SIMPLE(metrics) → ScoredDecision
//! achieved = (entropy in band) ∧ (max_func ≤ gate)
//! score    = min(per-metric qualities)   # reporting only; does not gate achieved
```

I mistook a pre-existing state for a change in this branch because I read the gate spec on the branch before diffing it against the base. There is no SIMPLE-semantics change here to split out, and no missing CHANGELOG entry for one.

**Consequence for #193:** it is stale, not open work — the fix landed in v0.4.3 and only lacked a CHANGELOG entry, which is why it stayed open. Recommend closing it as already resolved (commented there).

Everything else in my earlier comment stands: `.agents/AGENTS.md` still describes 3 generators and an 8-element lattice, and the `NAVIGABLE` thresholds are still `PROVISIONAL`.

Issue #239, opened 2026-07-25
macOS release build: non-deterministic duplicate-symbol link failure in liblbug.a


jeremy-wayland:
## What

`build (macos-latest, macos-arm64)` and `build-pypi-wheels (macos-latest, aarch64)` in `.github/workflows/release.yml` intermittently fail at the link step with:

```
duplicate symbol '...' in:
    .../liblbug.a[...]
    .../libantlr4_runtime.a[...]   (also seen: libantlr4_cypher.a, libzstd.a, libroaring_bitmap.a, libparquet.a, libminiz.a, libfastpfor.a)
ld: 7399 duplicate symbols
clang: error: linker command failed with exit code 1
error: could not compile `topos` (bin "topos") due to 1 previous error
```

`lbug` (`topos/engine/Cargo.toml`, pinned `=0.17.1`) bundles its own compiled copies of antlr4, zstd, roaring-bitmap, parquet, miniz, and fastpfor. Those same C++ libraries are also linked separately via their own Rust crates elsewhere in the dependency graph. Apple's linker rejects the resulting duplicate symbol definitions; GNU/Linux linkers tolerate it, which is why the Linux build jobs are unaffected.

## Why it's confusing

The failure is **non-deterministic given identical inputs**:

- Runner image: `macos-26-arm64`, version `20260715.0248.1` — identical across all runs below.
- Rust toolchain: `1.97.1 aarch64-apple-darwin` — identical.
- No build cache involved — all runs below show `No cache found.` (fresh, cold builds).
- `Cargo.lock` / `lbug` pin — identical across all runs below.

Evidence:
- v0.4.0 tag push (05:40, run [30146162292](https://github.com/Krv-Labs/topos/actions/runs/30146162292)) — macOS build **succeeded**.
- PR #223 `feat/clawhub-skill-card-governance` (06:11, run [30147066465](https://github.com/Krv-Labs/topos/actions/runs/30147066465)) — macOS build **failed**, identical `ld: 7399 duplicate symbols` signature. This PR touches only `scripts/check_skill.py` and `skills/topos/SKILL.md` — no Rust/Cargo changes at all.
- PR #238 (06:18, run [30147276724](https://github.com/Krv-Labs/topos/actions/runs/30147276724)) — both macOS jobs **failed**, same signature.

[...]

PR #284, opened 2026-08-04
ci: unbreak the rust job (rustfmt drift + missing liblbug)

jeremy-wayland:
Makes the `rust` CI job pass on `release/v0.4.4`, which is currently red on every run. Two independent causes, both pre-existing and neither specific to any feature branch.

## 1. `Format check` — rustfmt drift

`cargo fmt --all --check` fails on `release/v0.4.4` itself, in `topos/engine/src/graphs/graphify/object.rs` and `topos/mcp/src/schemas.rs`. CI installs `dtolnay/rust-toolchain@stable` and there is no `rust-toolchain.toml`, so a newer stable rustfmt reformats what an older one wrote. Nothing was misformatted on purpose.

This commit is rustfmt output only — no logic change. Local rustfmt 1.8.0 produces output byte-identical to what the failing job asked for.

## 2. `Clippy` — the job never had liblbug

With the format check passing, the next step fails:

```
error: could not find native static library `lbug`, perhaps an -L flag is missing?
error: could not compile `lbug` (lib) due to 1 previous error
```

Clippy and the test suite both link `topos-engine`, which depends on `lbug = "=0.17.1"` — so the `rust` job needs liblbug exactly as much as `composable` does. `composable` sets it up in three steps and passes; `rust` has none of them and fails at link time.

This adds those same three steps, copied from `composable` in the same file and in the same order: apt `cmake libssl-dev pkg-config`, cache `~/.cargo/lbug-prebuilt`, then `scripts/setup-lbug-prebuilt.sh` — whose header states its purpose is to point `build.rs` at a pinned prebuilt so it skips both the in-crate downloader and the cmake source fallback (issue #239).

### Why it went unnoticed

`Format check` runs before `Clippy`, so while cause 1 was failing, clippy never executed and cause 2 stayed invisible. Fixing the formatting is what surfaced it — confirmed deterministic across two runs of #278.

`main` has a byte-identical `rust` job and the same pinned `lbug`, and is green only because its cargo cache carries the built artifact. Any cache miss or eviction there fails the same way, so this hardens `main` too rather than only unblocking the release branch.

## Verification

- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test --workspace` (487 tests) all clean locally.
- `ci.yml` parses, and the `rust` job's step order now matches `composable`'s.
- The real check is this PR's own `rust` job: it must get past `Format check` and through `Clippy` and `Tests`, which is the first time those steps will have run on this branch family.

Unblocks the `rust` check on #278.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,11 @@ jobs:
     steps:
       - uses: actions/checkout@v7
 
+      - name: Install build deps (lbug / OpenSSL)
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends cmake libssl-dev pkg-config
+
       - name: Install Rust toolchain
         uses: dtolnay/rust-toolchain@stable
         with:
@@ -19,6 +24,19 @@ jobs:
 
       - name: Cache cargo
         uses: Swatinem/rust-cache@v2
+
+      # Clippy and the test suite link topos-engine, so this job needs
+      # liblbug just as much as `composable` does. Without it, a cold cargo
+      # cache fails at link time with "could not find native static library
+      # `lbug`". See scripts/setup-lbug-prebuilt.sh and issue #239.
+      - name: Cache lbug prebuilt
+        uses: actions/cache@v4
+        with:
+          path: ~/.cargo/lbug-prebuilt
+          key: lbug-prebuilt-0.17.1-static-${{ runner.os }}-${{ runner.arch }}
+
+      - name: Setup lbug prebuilt
+        run: ./scripts/setup-lbug-prebuilt.sh
 
       - name: Check versions (Cargo.toml is source of truth)
         run: python3 scripts/check_versions.py
--- a/topos/engine/src/graphs/graphify/object.rs
+++ b/topos/engine/src/graphs/graphify/object.rs
@@ -45,7 +45,10 @@ pub enum GraphifyError {
     /// Neither a `"links"` nor an `"edges"` top-level array was present.
     MissingEdgeKey,
     /// File exceeded [`MAX_GRAPH_JSON_BYTES`].
-    TooLarge { path: PathBuf, size: u64 },
+    TooLarge {
+        path: PathBuf,
+        size: u64,
+    },
 }
 
 impl std::fmt::Display for GraphifyError {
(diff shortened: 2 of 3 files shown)


sgathrid:
Closing as superseded by #278, which carries everything here plus the piece this was missing. The diagnosis in this PR was correct and is what led to the full fix — recording the detail so it is not rediscovered later.

## Everything here is already on `release/044-256-harness-install`

| Change | Status on #278 |
| --- | --- |
| rustfmt `topos/engine/src/graphs/graphify/object.rs` | byte-identical (arrived via `3a9fe94`) |
| rustfmt `topos/mcp/src/schemas.rs` | byte-identical (same) |
| `ci.yml`: apt build deps, lbug cache, lbug setup in the `rust` job | present, plus the fix below |

Verified by direct comparison against this branch's head (`5e1930c`): both Rust files diff empty, and #278's `rust` job is a strict superset of these steps.

## Why the extra fix was needed

The premise here is right: the `rust` job runs clippy and the test suite, both of which link `topos-engine`, but only `composable` installed liblbug.

The remedy is incomplete. lbug's build script emits the `-L` search path only when `LBUG_LIBRARY_DIR` is set, and it registers no `rerun-if-env-changed` for that variable. So a `target/` restored from a run predating the variable keeps its stale build-script output — no `-L` — and still fails to link no matter what the environment now says. Adding the setup step does not invalidate that cache.

**This was invisible because the failure only appears on a cold cargo cache, and no recent `rust` run had one.** The green tick on this PR came from a primary-key cache hit in which lbug was never recompiled, as did every prior green run on this job. The lbug path has effectively not been exercised in `rust` for as long as the cache has been warm.

#278 changed `Cargo.lock` (a tree-sitter-rust bump), which dropped the job to a restore-key fallback and surfaced it: the first push there still failed with

```
error: could not find native static library `lbug`, perhaps an -L flag is missing?
```

*after* these exact steps had been applied.

## What #278 does instead

Moves the lbug setup ahead of the cargo cache, so `LBUG_LIBRARY_DIR` is already exported when `rust-cache` computes its key, and adds that variable to `env-vars` so a cache built without lbug is never restored into a job that needs it.

The passing run compiled lbug three times from a fully cold cache (351 crates from scratch) — the first genuine end-to-end exercise of this path rather than a restore around it.

No action needed here; nothing is lost by closing. If #278 is rejected for unrelated reasons, this should be reopened **with** the ordering and `env-vars` change, not as-is.

sgathrid:
Superseded by #278 — see the comment above for the full accounting.

PR #353, opened 2026-08-24
fix(cli): discovery and path filtering robustness (#352, #334, #335)

### LINKED ISSUE #334: Discovery silently skips source directories named coverage/, build/, out/, dist/, env/
## Summary

`SKIP_DIR_NAMES` in [`topos/engine/src/adapters/discovery.rs:37`](../blob/main/topos/engine/src/adapters/discovery.rs#L37) matches directory names at **any depth**, so legitimate source directories named `coverage`, `build`, `out`, `dist`, or `env` are silently dropped from evaluation. No warning is emitted and the files never appear in the results, so the omission is invisible to the caller.

## Evidence

Found while benchmarking Topos against the [qlty](https://github.com/qltysh/qlty) codebase (947 files scanned, `topos evaluate . --recursive --json`).

`qlty-cli/src/commands/` contains both a module file and its directory:

```
qlty-cli/src/commands/coverage.rs          <- scanned
qlty-cli/src/commands/coverage/complete.rs <- SKIPPED
qlty-cli/src/commands/coverage/publish.rs  <- SKIPPED  (546 LOC)
qlty-cli/src/commands/coverage/transform.rs<- SKIPPED
qlty-cli/src/commands/coverage/utils.rs    <- SKIPPED
```

```console
$ jq -r '.results[].file' topos-on-qlty.json | grep 'commands/coverage'
./qlty-cli/src/commands/coverage.rs
```

`publish.rs` is one of the highest-complexity files in that repository (546 LOC, 15 functions, cyclomatic 83 per `qlty metrics`) — exactly the kind of file the SIMPLE and NAVIGABLE gates exist to catch. A competing tool run over the same tree analyzed all four files.

This is a correctness problem for a tool whose value proposition is rigor: a project rollup takes the **project-wide minimum per dimension**, so a silently skipped file can turn a failing verdict into a passing one.

## Affected names

Of the current `SKIP_DIR_NAMES` list, these are all plausible real source directory names in a normal Rust/Python/TS project:

- `coverage` (Rust module, as above)
- `build` (build-system source, e.g. `src/build/`)
- `out` , `dist` (rarer but legal module names)
- `env` (Python package named `env`, distinct from a virtualenv)

The rest (`.git`, `.gitnexus`, `node_modules`, `target`, `__pycache__`, `.venv`, cache dirs, …) are safe to keep matching unconditionally.

## Proposed fix

Make the ambiguous names conditional rather than absolute. Any one of these resolves it:

1. **Only skip at repo root** — `coverage/` at the top level is a report directory; `src/commands/coverage/` is a module. Cheapest and covers the real-world cases.
2. **Skip only if the directory contains no supported source files** (or only generated/HTML output). Slightly more expensive, most accurate.
3. **Defer to `.gitignore`** for the ambiguous names — build outputs are ignored, source is not.

Whichever route: emit a `warnings` entry listing skipped directories so the omission is at least visible in `--json` output and in the MCP response.

## Verification

Add a discovery test with a fixture tree containing `src/commands/coverage/publish.rs` and a root-level `coverage/index.html`, asserting the first is scanned and the second is not.

### LINKED ISSUE #335: No default exclusions for fixtures and generated code; .toposignore is opt-in and unused here
## Summary

Topos scores test fixtures, generated code, and deliberately-broken lint inputs as if they were hand-written source. There is no default exclusion and `.toposignore` is opt-in, so out-of-the-box project verdicts are dominated by files nobody wrote or maintains. Because the project rollup takes the **minimum per dimension**, a single generated file can floor the whole verdict.

## Evidence

`topos evaluate . --recursive` over the [qlty](https://github.com/qltysh/qlty) repository (947 files):

| Observation | Value |
| --- | --- |
| Files rated `SLOP` | 11 — **all** of them lint-plugin fixtures with intentionally invalid syntax (`wont_compile.rs`, `count_nested_if.go`, …) |
| Worst `ast.max_function_complexity` | 133 (`qlty.tests.v1.serde.rs`) and 118 (`qlty.analysis.v1.serde.rs`) — both prost/serde **generated** protobuf code |
| Worst `nav.max_function_divergence` | 145.6 (`visitor.rs`, real) vs 31.2 (`file_complexity.js` / `file_complexity.py`, both fixtures designed to be complex) |

For comparison, qlty's own config excludes exactly this class of file by default:

```toml
exclude_patterns = ["**/fixtures/**", "**/protos/**", "**/*.in/**", "plugins/**"]
```

and `qlty init` writes a ~25-entry starter `exclude_patterns` list (`**/generated/**`, `**/testdata/**`, `**/vendor/**`, `*.min.*`, …) plus `test_patterns` for every new repo.

Topos ships neither a default list nor a generated starter file — and this repo does not have a `.toposignore` of its own, so Topos does not dogfood the mechanism either.

## Proposed fix

1. **Ship a conservative default skip list** for unambiguous non-source: `**/*.min.*`, `**/vendor/**`, `**/testdata/**`, `**/__snapshots__/**`, `**/*.in/**`, and files whose first ~5 lines contain a `@generated` / `Code generated by ... DO NOT EDIT` marker. Keep it small and documented; anything ambiguous stays in.
2. **Have `topos config` write a starter `.toposignore`** (commented, with the common fixture/generated patterns) the way `qlty init` writes `exclude_patterns`.
3. **Report what was excluded** — a count in the rollup and the patterns that matched, so exclusions are auditable rather than silent.
4. **Add a `.toposignore` to this repository** covering `tests/fixtures/**` and `benchmarks/**`, so Topos's self-score reflects hand-written code.

## Why this matters

Two of Topos's four pillars gate on per-function maxima (`ast.max_function_complexity`, `nav.max_function_divergence`). Generated code reliably blows past both. Every user who points Topos at a real repository hits this on day one, and the failure mode — a `SLOP` verdict caused by a file the user does not own — reads as the tool being wrong rather than the corpus being wrong.

Related: #334 (silently skipped source directories). Both are about making the scanned set correct and visible.

### LINKED ISSUE #352: Fatal exit code when evaluating directory with 0 supported source files (type-only / config packages)
### What happened?
In monorepo workspaces containing configuration-only packages (e.g., `packages/typescript-config`) or pure type definition packages (e.g., `packages/compute-contracts` with only JSON/configs or unsupported stubs), running Topos on the directory exits with a fatal error.

### Root Cause
The directory evaluator returns a fatal error exit code if zero supported source files are found in the target directory.

### Expected Behavior / Proposed Enhancement
Gracefully skip directories/packages containing 0 supported source files with an informational message/note rather than returning a fatal exit code.

jeremy-wayland:
### PR 1: CLI Discovery & Path Filtering Robustness

Resolves #352, resolves #334, resolves #335.

#### Summary of Changes:
1. **Graceful exit on 0 supported source files (#352)**:
   - When evaluating directories containing zero supported source files (such as type-only packages or configuration-only packages),  now exits with status code `0` and emits an informational message (`info: no supported source files found...`), rather than a fatal error exit code `1`.
   - `--json` output returns `{"version": "...", "results": [], "warnings": []}` with exit code `0`.
   - Explicit named missing paths or unsupported file extensions still return fatal exit codes as expected.

2. **Scope ambiguous skip names to root only (#334)**:
   - Scoped ambiguous directory skip names (`coverage`, `build`, `out`, `dist`, `env`) to top-level/root traversal only.
   - Legitimate nested source modules like `src/commands/coverage/` or `src/build/` are now fully scanned.
   - Common non-source/virtualenv directory names (`node_modules`, `.venv`, `target`, `__pycache__`, etc.) and actual virtualenvs continue to be skipped unconditionally at any depth.

3. **Built-in ignore defaults and `.toposignore` (#335)**:
   - Added conservative built-in ignore defaults for minified files (`*.min.*`), test fixtures/vendor/testdata/snapshots directories (`fixtures`, `vendor`, `testdata`, `__snapshots__`), and files starting with `@generated` / `Code generated by ... DO NOT EDIT` header markers in their first lines.
   - Added repository root `.toposignore` excluding `tests/fixtures/`, `benchmarks/`, and `scratch/`.

#### Verification:
- All unit, integration, and doc tests pass (`cargo test`).
- Verified with `cargo clippy --all-targets` and `cargo fmt --check`.
- Ran discovery test cases for nested modules, root skips, and generated header exclusions.


jeremy-wayland (commented):
LGTM! Fixes discovery edge cases (#352, #334, #335) cleanly with comprehensive test coverage.

Issue #350, opened 2026-08-24
Contradictory ✓ PASS displayed in CLI output when diagnostic score is 0%


jeremy-wayland:
### Summary
CLI evaluation outputs can display contradictory statuses such as `COMPOSABLE ✓ PASS 0%`.

### Root Cause
The gate status checkmark evaluates binary policy criteria independently of the underlying continuous diagnostic score, producing confusing output where a pillar passes binary gating despite a 0% metric score.

### Proposed Enhancement
Improve formatting when binary policy passes but the diagnostic score is critically low:
- Display `~ PASS (LOW SCORE)` or `! WARN` when the diagnostic score falls below 25%.

Issue #345, opened 2026-08-23
Support `pi`


jeremy-wayland:
### What do you want?

When running `topos install` can we also install for the pi harness

### Why?

Have been using pi quite a bit and would be great if we easily hooked into this as well.

sgathrid:
Closed via #368

PR #355, opened 2026-08-24
feat(cli): CLI determinism, status clarity & pi harness support

### LINKED ISSUE #332: CLI --json key order is nondeterministic run to run, and disagrees with MCP for the same data
Surfaced while verifying #322 / [#331](https://github.com/Krv-Labs/topos/pull/331) — A/B-diffing `topos evaluate --json` across two binaries produced ~20k characters of diff in which **every value was identical** and only key ordering moved. That made a routine equivalence check expensive to interpret.

## What

Two runs of the *same* `topos` binary on the same input produce byte-different JSON:

```
cargo run --release -q --bin topos -- evaluate <corpus> -r --json --no-composable > r1.json
cargo run --release -q --bin topos -- evaluate <corpus> -r --json --no-composable > r2.json

diff r1.json r2.json          # differs
diff <(jq -S . r1.json) <(jq -S . r2.json)   # identical
```

## Why

`ClassificationResult` holds its maps as `HashMap` (`topos/engine/src/core/characteristic_morphism.rs:69-79` — `dimensions`, `scores`, `raw_metrics`, `interpretation`). Rust's `HashMap` uses `RandomState`, so iteration order is randomized per process.

That alone would normally be invisible, because `serde_json`'s default `Map` is `BTreeMap`-backed and re-sorts on the way out. But `topos/cli/Cargo.toml:27` enables `serde_json`'s `preserve_order` feature:

```toml
# `preserve_order` keeps harness config keys in their original order when
serde_json = { version = "1.0", features = ["preserve_order"] }
```

`preserve_order` swaps that `BTreeMap` for an `IndexMap`, which faithfully preserves *insertion* order — and insertion order here comes straight from the randomized `HashMap` iteration. So the feature that exists to stabilize harness-config round-tripping is what de-stabilizes metric output.

## The part that is easy to miss

Cargo features are additive and unified per build, and `serde_json` is a single shared dependency. Confirmed with `cargo tree -e features`:

- `cargo tree -p topos` → `serde_json feature "preserve_order"`, `"indexmap"`
- `cargo tree -p topos-mcp` → only `"default"`, `"std"`, `"alloc"`

So **the MCP server and the CLI serialize the same `ClassificationResult` with different key ordering**: MCP is sorted and deterministic, the CLI is randomized. Same struct, same data, different bytes, purely from which binary pulled the feature in.

## Impact

Not a correctness bug — no value is ever wrong. It costs elsewhere:

- Any golden-file or snapshot test over CLI `--json` cannot be byte-compared without normalizing first.
- Before/after diffing of `--json` (exactly what an agent does to check whether a refactor changed anything) is drowned in ordering noise.
- Response bytes are unhashable/uncacheable across runs.
- CLI and MCP outputs cannot be compared to each other directly.

## Suggested fix

Prefer `BTreeMap<String, _>` for the four `ClassificationResult` maps. That makes ordering deterministic and sorted **regardless of which crate's feature flags win**, rather than leaving correctness of the output format dependent on Cargo feature unification. The maps are small (a dozen-odd metrics), so the lookup difference is irrelevant here, and `preserve_order` keeps doing its intended job for harness config.

Worth checking whether any other serialized struct carries a `HashMap` with the same exposure before settling on the shape.

## Verification

A regression test running the same evaluation twice in-process and asserting byte-identical serialization would pin this — and would fail today only for the CLI's feature set, so it should live where `preserve_order` is active.

jeremy-wayland:
### Summary

Resolves #332, #350, #345.

1. **Byte-deterministic JSON serialization (#332):** Converted `ClassificationResult` maps (`dimensions`, `scores`, `raw_metrics`, `interpretation`) and `combine_dimensions` to `BTreeMap` so JSON outputs preserve alphabetical key ordering.
2. **Status clarity for low scores (#350):** Updated the summary table renderer so passing pillars with diagnostic scores below 25% display `! WARN` rather than contradictory `✓ PASS 0%`.
3. **Pi harness configuration (#345):** Added `pi` to supported harnesses configuring `~/.pi/agent/settings.json` with detection on `~/.pi`.

### DIFF
--- a/topos/cli/src/commands/evaluate/info.rs
+++ b/topos/cli/src/commands/evaluate/info.rs
@@ -318,7 +318,7 @@ fn terminal_width(term: &Term, fallback: usize) -> usize {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::HashMap;
+    use std::collections::BTreeMap;
 
     use topos_engine::core::omega::EvaluationValue;
 
@@ -327,7 +327,7 @@ mod tests {
     fn result(simple: f64, secure: f64) -> ClassificationResult {
         ClassificationResult {
             is_parseable: true,
-            scores: HashMap::from([
+            scores: BTreeMap::from([
                 ("simple".to_string(), simple),
                 ("secure".to_string(), secure),
             ]),
--- a/topos/cli/src/commands/evaluate/info_render.rs
+++ b/topos/cli/src/commands/evaluate/info_render.rs
@@ -239,7 +239,7 @@ fn weakest_dimension(result: &ClassificationResult) -> (&'static str, f64) {
 
 #[cfg(test)]
 mod tests {
-    use std::collections::HashMap;
+    use std::collections::{BTreeMap, HashMap};
 
     use topos_engine::core::omega::EvaluationValue;
 
@@ -247,7 +247,7 @@ mod tests {
 
     fn result(simple: f64, secure: f64) -> ClassificationResult {
         ClassificationResult {
-            scores: HashMap::from([
+            scores: BTreeMap::from([
                 ("simple".to_string(), simple),
                 ("secure".to_string(), secure),
             ]),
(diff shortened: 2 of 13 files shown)


jeremy-wayland (commented):
LGTM! Deterministic serialization (#332), low-score status warning (#350), and pi harness support (#345) verified.

PR #360, opened 2026-08-24
feat(cli): CLI determinism, status clarity & pi harness support

### LINKED ISSUE #332: CLI --json key order is nondeterministic run to run, and disagrees with MCP for the same data

jeremy-wayland:
### Summary

Resolves #332, #350, #345.

1. **Byte-deterministic JSON (#332)** — `ClassificationResult` *and* every MCP wire map in `schemas.rs` are `BTreeMap`, so `topos evaluate --json` and every MCP structured payload emit sorted keys. This matters because the CLI builds `serde_json` with `preserve_order` and links `topos-mcp`, so feature unification makes `serde_json::Map` an insertion-ordered `IndexMap` for the whole build — a `HashMap` field serialized in `RandomState` order, differently on every process. Converting the wire types let four `BTreeMap -> HashMap` conversion blocks collapse to `.clone()` and deleted `formatting::sorted_pairs`.
2. **Status clarity for low scores (#350)** — a passing pillar whose score is below `WEAK_SCORE` (25%) renders `! WARN` instead of a contradictory `✓ PASS 0%`. In the multi-file card the status keys off the *weakest* file, since that row prints AVG and MIN side by side. Gates alone still decide pass/fail; the threshold feeds no verdict or exit code and `--json` is unchanged.
3. **pi harness (#345)** — ninth harness, with **two** artifacts. See below.

### pi: no MCP client, so one artifact was not enough

pi has no MCP client of its own — its README says *"No MCP. […] build an extension that adds MCP support"*, and `settings-manager.ts` declares `packages`, `extensions` and `skills` with **zero** occurrences of `mcp`. Two consequences:

**The config path was wrong.** `~/.pi/agent/settings.json` is pi's own settings file (theme, provider, transport); an `mcpServers` entry there is read by nothing, so `topos status` would have reported `✓ active` for a registration pi never sees. The entry now goes to `~/.pi/agent/mcp.json`, which the [`pi-mcp-adapter`](https://github.com/nicobailon/pi-mcp-adapter) extension reads. Of the six locations that adapter searches, it is the only one that is pi's alone — `~/.config/mcp/mcp.json` and `~/.agents/mcp.json` rank higher but are shared across tools and are not a per-harness installer's to own.

**An MCP entry alone configures nothing on pi**, so `topos install pi` also writes the route that works today: a reference to an already-installed topos skill directory in `settings.json`'s `skills` array (a documented array of paths whose example in pi's own docs is other agents' skill directories). topos still never writes skill content — it appends one directory path; skills stay ClawHub / Hermes / openclaw's, as `residue.rs` requires. Three outcomes, kept distinct because collapsing them gives wrong advice:

| Skill found | Install writes |
| --- | --- |
| in `~/.pi/agent/skills` — pi's default scan, including openclaw's symlink farm into `~/.agents/skills` | nothing; reported `discovered` |
| in `~/.agents/skills` or `~/.claude/skills` | that directory, appended to `skills` |
| nowhere | nothing; names `openclaw skills install @Krv-Labs/topos` |

The two artifacts are never folded together: a missing skill must not mask a working MCP entry, nor the reverse, so the reference gets its own status line and its own `skillRef` object in `status --json`. pi also carries the second `note` (after Antigravity), shown on install *and* in `status`: the MCP entry is inert until the adapter is installed.

### Commits

| | |
| --- | --- |
| `6f2cdfd` | wire maps -> `BTreeMap`; drop the four conversions and `sorted_pairs` (**-22 lines net**) |
| `4ce4521` | sorted-key regression guards on both JSON channels |
| `d53ebe3` | name `WEAK_SCORE`; fix the card caption that still claimed status was gates-only |
| `f6c5726` | WARN keys off the weakest file, not the project average |
| `a266fe4` | pi writes to `mcp.json`, not `settings.json`; adapter note |
| `3db1a9f` | ADR harness matrix + README checklist |
| `3b13d3d` | CHANGELOG `[Unreleased]` entries |
| `b542f9d` | pi's skill reference in `settings.json` — new `skills_entry` module, 3 e2e tests |
| `244da9f` | ADR: pi's second artifact and why `settings.json` has no MCP key |

### Verification

`cargo test --workspace` (634 tests), `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `check_skill.py`, `check_agent_plugin.py` — all clean. End-to-end against the built binary:

```
$ topos evaluate ... --json
scores      SORTED  ['navigable', 'secure', 'simple']
raw_metrics SORTED  ['ast.entropy', 'ast.max_function_complexity', 'cfg.cyclomatic', ...]
```

`cargo tree` confirms `serde_json feature "indexmap"` is enabled workspace-wide, so `preserve_order` really is on for the shipped binary.

Against a real pi install with the openclaw symlink farm:

```
$ topos install pi --dry-run
│  pi
│    ○ [dry run] would register the MCP server in ~/.pi/agent/mcp.json
│    ✓ topos skill already discovered in ~/.pi/agent/skills — no reference needed
│    ! pi reads MCP servers only through its adapter extension — run
│      `pi install npm:pi-mcp-adapter` if you have not already, or this entry is inert.
```

The uninstall round trip is snapshot-asserted: no file and no directory left behind, and the skill itself untouched.


sgathrid (changes requested):
## Requesting changes: two issues — one silent data loss, one small JSON shape change

### 1. Uninstall removes what it didn't add — must fix

`skills_entry.rs:151-164` removes any `.agents/skills` / `.claude/skills` entry with no ownership check.

If the user hand-added `"skills": ["~/.agents/skills"]` before ever running `topos`, install is a no-op (`Active` → `Ok(None)`, nothing recorded) — but uninstall still deletes it.

**Fix:** only remove the entry if install actually appended it.

### 2. `status --json` adds `"skillRef": null` to 8 harnesses — fix

`status.rs:195` uses `.then(...)`, so `None` becomes `null`. Omit the key entirely when `skill_ref == false`.

jeremy-wayland:
Release cut: **#367** (`release/v0.6.0`) merges this branch and bumps to v0.6.0. Prefer merging/closing via that release PR to avoid duplicate commits on `main`.

jeremy-wayland:
Release plan updated: merge this to `main` first. The release branch/PR should then be **version bumps only** (`CHANGELOG` → `[0.6.0]`, `Cargo.toml` / registry / skill / extension → `0.6.0`) — no feature merge on the release branch.

